PR #17, opened 2018-09-19
PLANNER-1274 Display route length and number of location

### LINKED ISSUE #2: Load best route after refresh
Currently, after refreshing browser, the list of locations is loaded in the order they are stored in the Spring repository. To solve this, simply ask solver for best solution (if solver has already been started) upon subscription.

### LINKED ISSUE #4: Display route length

yurloc:
Fixes #2, fixes #4.

### DIFF
--- a/optaweb-tsp-planner/src/main/java/org/optaweb/tsp/optawebtspplanner/RouteMessage.java
+++ b/optaweb-tsp-planner/src/main/java/org/optaweb/tsp/optawebtspplanner/RouteMessage.java
@@ -1,0 +1,22 @@
+package org.optaweb.tsp.optawebtspplanner;
+
+import java.util.List;
+
+public class RouteMessage {
+
+    private final String distance;
+    private final List<Place> route;
+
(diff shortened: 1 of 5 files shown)


yurloc:
@rsynek I hope to find an answer for the first question next sprint: https://issues.jboss.org/browse/PLANNER-1276.

Second question - can you be more specific? I assume we don't want to test that what's been sent is what I receive on client - that is the job of the underlying libraries and frameworks that handle network communication and should have their own test coverage.

rsynek:
@yurloc truth, I am thinking of an integration test covering client logic which is feed by data sent from a server instead of mock data prepared by a test on the client side (this might be an outcome of PLANNER-1276).

We should definitely focus on client unit coverage and server unit coverage, still it would be helpful to have a small number of integration tests binding client & server together. Imagine we upgrade some of the libraries in the stack - a test demonstrating it works all together gives us a better level of certainty that no incompatibility was introduced.

PR #19, opened 2018-10-03
PLANNER-1299

danielefiungo:
One mvn build that delegates to the frontend build

### DIFF
--- a/optaweb-tsp-ui/src/Location.test.js
+++ b/optaweb-tsp-ui/src/Location.test.js
@@ -23,23 +23,16 @@ describe('Location Component', () => {
     const props = {
       id: 10,
       removeDisabled: false,
-      removeHandler: jest.fn(() => {}),
-      selectHandler: jest.fn(select => select),
+      removeHandler: jest.fn(),
+      selectHandler: jest.fn(),
     };
     expect.assertions(2);
-    const location = shallow(
-      <Location
-        id={props.id}
-        removeDisabled={props.removeDisabled}
-        removeHandler={props.removeHandler}
-        selectHandler={props.selectHandler}
-      />,
-    );
+    const location = shallow(<Location {...props} />);
     expect(location).toMatchSnapshot();
 
     location.find('button').simulate('click');
     location.find('div').simulate('mouseEnter');
 
-    expect(props.removeHandler.mock.calls.length).toBe(1);
+    expect(props.removeHandler).toHaveBeenCalledTimes(1);
   });
 });
--- a/optaweb-tsp-ui/src/LocationList.test.js
+++ b/optaweb-tsp-ui/src/LocationList.test.js
@@ -24,9 +24,9 @@ describe('Location List Component', () => {
       route: [],
       domicileId: 1,
       distance: '10',
-      removeHandler: jest.fn(() => {}),
-      selectHandler: jest.fn(select => select),
-      loadHandler: jest.fn(() => {}),
+      removeHandler: jest.fn(),
+      selectHandler: jest.fn(),
+      loadHandler: jest.fn(),
     };
     expect.assertions(2);
     const locationList = shallow(<LocationList {...props} />);
@@ -34,7 +34,7 @@ describe('Location List Component', () => {
 
     locationList.find('button').simulate('click');
 
-    expect(props.loadHandler.mock.calls.length).toBe(1);
+    expect(props.loadHandler).toHaveBeenCalledTimes(1);
   });
 
   it('should render correctly with a few routes', () => {
@@ -58,16 +58,15 @@ describe('Location List Component', () => {
       ],
       domicileId: 1,
       distance: '10',
-      removeHandler: jest.fn(() => {}),
-      selectHandler: jest.fn(select => select),
-      loadHandler: jest.fn(() => {}),
+      removeHandler: jest.fn(),
+      selectHandler: jest.fn(),
+      loadHandler: jest.fn(),
     };
     expect.assertions(2);
     const locationList = shallow(<LocationList {...props} />);
     expect(locationList).toMatchSnapshot();
 
-    expect(locationList.find('Location').length).toBe(props.route.length);
-
+    expect(locationList.find('Location')).toHaveLength(props.route.length);
   });
 
 });
--- a/optaweb-tsp-ui/src/TspMap.test.js
+++ b/optaweb-tsp-ui/src/TspMap.test.js
@@ -44,8 +44,8 @@ describe('TSP Map View', () => {
       ],
       domicileId: 1,
       distance: '10',
-      clickHandler: jest.fn(() => {}),
-      removeHandler: jest.fn(() => {}),
+      clickHandler: jest.fn(),
+      removeHandler: jest.fn(),
       zoom: 5,
       selectedId: 1
     };


.gitignore
```diff
@@ -59,3 +59,16 @@ fabric.properties
...
+.classpath
+.factorypath
+.project
+.settings/
+.springBeans
+.sts4-cache
+
+
+
+
```
ge0ffrey: code style (soft suggestion): max 2 empty lines in general, here'd just leave 1

.idea/misc.xml
```diff
@@ -7,6 +7,7 @@
     <option name="originalFiles">
       <list>
         <option value="$PROJECT_DIR$/optaweb-tsp-planner/pom.xml" />
+        <option value="$PROJECT_DIR$/optaweb-tsp-ui/pom.xml" />
```
ge0ffrey: @yurloc .idea files shouldn't be checked in. Please use the same .gitignore file as the optaplanner repository.

Please use an *.adoc for the readme, like the he optaplanner repository.
Please use the same license file with the same name, like in the optaplanner repository's root dir.

optaweb-tsp-planner/pom.xml
```diff
@@ -1,65 +1,107 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
+    <modelVersion>4.0.0</modelVersion>
```
ge0ffrey: @danielefiungo please set your xml indentation to 2 spaces.
We have Intellij/Eclipse code style files here:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/master/ide-configuration

optaweb-tsp-planner/pom.xml
```diff
@@ -1,65 +1,107 @@
...
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-websocket</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.optaplanner</groupId>
+            <artifactId>optaplanner-examples</artifactId>
+            <version>7.12.0-SNAPSHOT</version>
```
ge0ffrey: @yurloc Do it similar like optaweb-employee-rostering:
https://github.com/kiegroup/optaweb-employee-rostering/blob/master/pom.xml

Itntroduce a maven property version.org.optaplanner, use the optaplanner-bom in dependencyManagement, etc Applies to all dependencies.

optaweb-tsp-planner/pom.xml
```diff
@@ -1,65 +1,107 @@
...
+        <profile>
+            <id>include-ui</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>org.optaweb.tsp</groupId>
+                    <artifactId>optaweb-tsp-ui</artifactId>
+                    <version>0.0.1-SNAPSHOT</version>
```
ge0ffrey: Use ${project.version} to avoid a pita during MichaelB's releasing (he needs to upgrade the version numbers then).
Better yet, all internal GAV's should be in the depenency management, so having to declare a `<version>` element to get begin with isn't needed. See optaplanner-examples pom.xml in optaplanner repo.

optaweb-tsp-planner/pom.xml
```diff
@@ -1,65 +1,107 @@
...
+                            <excludeTransitive>true</excludeTransitive>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+    <repositories>
+        <repository>
+            <id>jboss-snapshot</id>
```
ge0ffrey: Only the root pom.xml should have a repositories element. That should be exactly the same as the one in optaweb-employee-rostering.

optaweb-tsp-ui/.gitignore
```diff
@@ -1,5 +1,6 @@
 # See https://help.github.com/ignore-files/ for more about ignoring files.
-
+### Maven Frontend Plugin ###
+node/
```
ge0ffrey: @danielefiungo Can we put the node output under the target directory?

optaweb-tsp-ui/package-lock.json
```diff
@@ -4,6 +4,12 @@
   "lockfileVersion": 1,
```
ge0ffrey: Should package-lock.json files be checked in? Seems like something that should be gitignored (preferably because its in the target directory, but if that's painful, we can add it to gitignore directly).

optaweb-tsp-ui/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>org.optaweb.tsp</groupId>
+	<artifactId>optaweb-tsp-ui</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
+	<name>optaweb-tsp-ui</name>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
```
ge0ffrey: This pom should extend the root pom. Similar to how optaweb-employee-rostering-shared has a parent optaweb-employee-rostering. When that parent is there, all these properties should be moved there, see the root pom of optaweb-employee-rostering

optaweb-tsp-ui/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+					it through HTTP as static content -->
+				<targetPath>static</targetPath>
+			</resource>
+		</resources>
+		<finalName>optaweb-tsp-ui</finalName>
+		<plugins>
+			<plugin>
+				<groupId>com.github.eirslett</groupId>
+				<artifactId>frontend-maven-plugin</artifactId>
+				<version>${version.frontend-maven-plugin}</version>
```
ge0ffrey: version goes in pluginManagement in root pom, see optaweb-employee-rostering root pom

pom.xml
```diff
@@ -0,0 +1,21 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.optaweb.tsp</groupId>
+    <artifactId>optaweb-tsp-parent</artifactId>
```
ge0ffrey: artfictId of any pom should match the directory name (unless there's a good reason not to do that), let's name this optaweb-traveling-salesman.

pom.xml
```diff
@@ -0,0 +1,21 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.optaweb.tsp</groupId>
```
ge0ffrey: @yurloc Does this match the java packages? Is it similar in style to how optaweb-employee-rostering does it?

optaweb-tsp-ui/package-lock.json
danielefiungo: Actually it should be committed because:

> - Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> - Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
> - To facilitate greater visibility of tree changes through readable source control diffs.
> - And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.
[npm docs](https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md#description)

Otherwise ignoring it allow us to always test the latest version of dependencies (especially in CI).

optaweb-tsp-planner/pom.xml
danielefiungo: I've applied the coding style. 

**Warning**: the create-react-app project came with eslint configuration that imply a code format.
In our case (tab-size) this don't create problems 2 space match with eslint style, but some other rule can clash with kiegroup standard code format, in this case the "eject" can be used to address the problem with all the related downsides.

optaweb-tsp-ui/.gitignore
danielefiungo: Yes, I'll add the required config

optaweb-tsp-planner/pom.xml
danielefiungo: I've found a rule conflict,
file : .js  rule: indent  **ESLint CRA** 2 spaces vs **KIE Conventions** 4 spaces

I've edited .eslint.json to adopt KIE Convention but this can became tricky to maintain and adapt.

yurloc: @danielefiungo KIE Conventions are Java-centric with the default indentation of 4 spaces. Special code style requirements for other languages need to override the default indentation. For example XML language overrides indentation to 2 spaces. I think we need to do the same for JavaScript and TypeScript as the standard indentation for JavaScript code seems to be 2 spaces. Most of the code samples I've seen so far use 2 spaces indentation [1]. What's your opinion, Daniele?

[1] @tiagobento uses 2 space indentation in his proof-of-concept app too: https://github.com/tiagobento/appformer-js/blob/master/packages/core/src/internal/Root.tsx

.idea/misc.xml
yurloc: I used the default flavor of [.gitignore for IDEA](https://www.gitignore.io/api/intellij). This flavor only ignores files under`.idea/` dir that are never intended to be version-controlled. There are other flavors that ignore [.iml files](https://www.gitignore.io/api/intellij+iml) too or [everything under .idea/ dir](https://www.gitignore.io/api/intellij+all) as well.

I thought it might be convenient for other developers that use IDEA to get the project up and running if the basic IDEA configuration is included in the repository. Some people are convinced that the repository [should be kept clean](https://github.com/joeblau/gitignore.io/issues/186#issuecomment-249319722) of **any IDE**-specific metadata, which I think is a legitimate approach. If that's your opinion too and unless you change your mind, I'm going to remove and ignore the whole `.idea/` directory.

optaweb-tsp-planner/pom.xml
yurloc: Yes, but out of scope of this PR (note taken).

optaweb-tsp-ui/package-lock.json
yurloc: AFAIU pacakge-lock.json is necessary for reproducible builds and it should be version-controlled.

optaweb-tsp-planner/pom.xml
yurloc: @danielefiungo `.eslint.json` is the single source of truth. It's based on airbnb style with just a few necessary tweaks. IDEA style need to be adjusted to produce code that has no ESLint errors. That's basically the option 2 that you have proposed. Do you agree with this approach?

optaweb-tsp-planner/pom.xml
```diff
@@ -54,12 +55,43 @@
...
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>org.optaweb.tsp</groupId>
+          <artifactId>optaweb-tsp-ui</artifactId>
+          <version>${project.version}</version>
+        </dependency>
+      </dependencies>
+    </profile>
+  </profiles>
```
yurloc: @danielefiungo What is the reason for having the dependency in a profile?

danielefiungo: The day UI can be deployed on a simple http server and include static resources inside BE can be optionally disabled in CI when UI is deployed separately.

@yurloc Make sense for you?

optaweb-tsp-ui/package-lock.json
```diff
@@ -10770,6 +11102,12 @@
         }
       }
     },
+    "underscore": {
+      "version": "1.4.4",
+      "resolved": "https://registry.npmjs.org/underscore/-/underscore-1.4.4.tgz",
+      "integrity": "sha1-YaajIBBiKvoHljvzJSA88SI51gQ=",
+      "dev": true
+    },
```
yurloc: @danielefiungo Can you explain these changes in `package-lock.json`? If I run `mvn clean install` locally, they get remove.

danielefiungo: Probably is related to test implementation I've add:
"enzyme": "^3.6.0" and "enzyme-adapter-react-16" without commit package.json to keep it separate. 
I'll revert it.

optaweb-tsp-planner/pom.xml
```diff
@@ -54,12 +55,43 @@
...
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>unpack-dependencies</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/classes</outputDirectory>
+              <excludeTransitive>true</excludeTransitive>
+            </configuration>
+          </execution>
+        </executions>
```
yurloc: @danielefiungo This has an unpleasant side effect. It unpacks all of this project's dependencies (including Spring Boot starters, H2 database, optaplanner examples) producing a sort of fat JAR. Some resources get overridden (e.g logback.xml). Can you configure the plugin to only unpack the `optaweb-tsp-ui` dependency?

optaweb-tsp-ui/package-lock.json
yurloc: Yeah, makes sense. Thank you!

optaweb-tsp-planner/pom.xml
danielefiungo: I've created and removed a review by mistake, anyway ...
the configuration prevent a minor side effect related to the unpack of UI jar that can be addressed with a mvn clean,  is minor compared your suggested side effects

package-lock.json
```diff
@@ -0,0 +1,551 @@
+{
```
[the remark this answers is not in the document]
danielefiungo: Sorry, my bad, probably I had run npm ... on the wrong directory.

PR #25, opened 2018-11-05
WIP: Typescript demo

danielefiungo:


optaweb-tsp-ui/package.json
```diff
@@ -34,12 +40,19 @@
...
+    "react-test-renderer": "^16.6.0",
+    "redux-devtools-extension": "^2.13.5",
+    "redux-mock-store": "^1.5.3"
+  },
+  "browserslist": [
+    ">0.2%",
+    "not dead",
+    "not ie <= 11",
+    "not op_mini all"
+  ]
```
yurloc: @danielefiungo What is the effect of this configuration and how did you come up with the values? I think we need to document it so that following questions can be answered:
* What is the purpose of this piece of code?
* Can we remove it? / Under what conditions can it be removed?
* Can we change the values? / When do we need to update it?
A simple link to the related package documentation or an article explaining the usage might be helpful.

PR #26, opened 2018-11-06
Simplify lint script

yurloc:


optaweb-tsp-ui/package.json
```diff
@@ -25,7 +25,7 @@
     "deploy": "gh-pages -d build",
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "lint": "./node_modules/eslint/bin/eslint.js ./src",
+    "lint": "npx eslint ./src",
```
danielefiungo: In the npx [docs](https://www.npmjs.com/package/npx):

> Executes <command> either from a local node_modules/.bin, **or from a central cache**, installing any packages needed in order for <command> to run.

Have you seen it in other projects? 
My only concern is : The central cache can override the eslint declared in package.json?

yurloc: Good question. My guess is that this is not our case because eslint is guaranteed to be in `node_modules/` since it's a dependency of react-scripts.

Central cache is `~/.npm/`?

yurloc: It seems that npx is commonly used in package.json scripts: https://github.com/search?p=1&q=npx+eslint+extension%3Ajson+size%3A%3C1000&type=Code.

danielefiungo: The docs don't specify where is the cache and even locally this is not a big problem, but
inside a container or compiling into a shared runner? 
I've made a quick search but nothing related to npx use in package.json probably the previous solution avoid this problem. 
At the end of the day you will type _npm run lint_ in both cases and npx is an handy tool for developers.

> It seems that npx is commonly used in package.json scripts: https://github.com/search?p=1&q=npx+eslint+extension%3Ajson+size%3A%3C1000&type=Code.

Code hits: 748 but the counterpart is 157K https://github.com/search?utf8=%E2%9C%93&q=.%2Fnode_modules%2Feslint%2Fbin%2Feslint.js+extension%3Ajson&type=Code&ref=advsearch&

yurloc: OK. Good point with the counterpart search.

Although I don't fully understand what the risk of using npx in scripts is, I believe you that it's safer to avoid (if I understood correctly).

So let's revert the `package.json` change and only update the README, right?

danielefiungo: Yes probably I've only some doubt and the safer way can be a better option.

PR #31, opened 2018-11-13
PLANNER-1289 Road distances

yurloc:


optaweb-tsp-planner/pom.xml
```diff
@@ -52,6 +52,11 @@
       <artifactId>spring-boot-starter-test</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.graphhopper</groupId>
+      <artifactId>graphhopper-reader-osm</artifactId>
+      <version>0.11.0</version>
```
rsynek: Why not introducing dependencyManagement already?

optaweb-tsp-planner/src/main/java/org/optaweb/tsp/optawebtspplanner/TspPlannerComponent.java
```diff
@@ -145,22 +183,40 @@ public void bestSolutionChanged(BestSolutionChangedEvent<TspSolution> bestSoluti
...
+                    location.getLongitude(),
+                    toLocation.getLatitude(),
+                    toLocation.getLongitude());
+            map.put(toLocation, graphHopper.route(there).getBest().getDistance());
+            GHRequest back = new GHRequest(
+                    toLocation.getLatitude(),
+                    toLocation.getLongitude(),
+                    location.getLatitude(),
+                    location.getLongitude());
+            // TODO handle no route -> roll back the problem fact change
```
rsynek: sounds like a good candidate for a test

yurloc: Yes, but since I'm not doing TDD, the code is untestable like this so I need to refactor it first. But I intend to do that as part of this PR.

optaweb-tsp-planner/pom.xml
yurloc: It looks like `<dependencyManagement>` only makes sense in a parent POM to manage childrens' dependencies: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management.

PR #32, opened 2018-11-22
Update optaplanner branding

ge0ffrey:


optaweb-tsp-ui/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "license": "Apache-2.0",
-  "homepage": "https://yurloc.github.io/optaweb-traveling-salesman/",
+  "homepage": "https://www.optaplanner.org",
```
yurloc: This field is not project metadata. The value is used when the app is served from a [relative path](https://github.com/facebook/create-react-app/blob/v2.0.4/packages/react-scripts/template/README.md#building-for-relative-paths), and not the server root. We might need to update it when the app will be deployed but it's probably going to be a different value. So this change doesn't make sense and I've excluded it.

PR #38, opened 2018-12-04
PLANNER-1289: Use GraphHopper to calculate road distances

yurloc:


yurloc:
@rsynek Added instructions for running the Belgium demo locally. Please check out and submit review.

README.adoc
```diff
@@ -5,3 +5,59 @@ image:https://travis-ci.org/kiegroup/optaweb-vehicle-routing.svg?branch=master[
...
++
+```shell
+dnf install npm
+```
+
+. Run frontend.
++
+```shell
+cd optaweb-tsp-ui
+npm start
```
rsynek: There should be `npm install` first

README.adoc
```diff
@@ -5,3 +5,59 @@ image:https://travis-ci.org/kiegroup/optaweb-vehicle-routing.svg?branch=master[
...
+
+. Download OpenStreetMap data for Belgium from Geofabrik.
+.. Open http://download.geofabrik.de/europe/belgium.html.
+.. Download the PBF file http://download.geofabrik.de/europe/belgium-latest.osm.pbf[belgium-latest.osm.pbf].
+.. Save it to `optaweb-vehicle-routing/optaweb-tsp-planner/local/belgium-latest.osm.pbf`.
+
+. Run backend.
++
+```shell
+./mvnw spring-boot:run
```
rsynek: Missing cd optaweb-tsp-planner

optaweb-tsp-planner/pom.xml
```diff
@@ -52,9 +52,43 @@
...
+        <plugin>
+          <groupId>org.pitest</groupId>
+          <artifactId>pitest-maven</artifactId>
+          <version>1.4.2</version>
+          <configuration>
+            <reportsDirectory>local/pit-reports</reportsDirectory>
+            <timestampedReports>true</timestampedReports>
+            <timeoutConstant>1000</timeoutConstant>
+            <mutators>
+              <!--
```
rsynek: May I ask why have been these mutators commented out?

optaweb-tsp-planner/src/main/java/org/optaweb/tsp/optawebtspplanner/TspMapController.java
```diff
@@ -58,7 +58,7 @@ public void create(Place place) {
     public void demo() {
 //        repository.findAll().forEach(planner::removePlace);
 //        repository.deleteAll();
```
rsynek: I know it's not a part of this PR, but if these lines of code are not used, let's erase them

optaweb-tsp-planner/src/test/java/org/optaweb/tsp/optawebtspplanner/OptawebTspPlannerApplicationTests.java
```diff
@@ -43,6 +44,7 @@
 @RunWith(SpringRunner.class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 @AutoConfigureMockMvc
+@ActiveProfiles("test")
```
rsynek: In general, String dependency does not provide any type safety. Would it be possible to define a constant and use it in all places we refer to the "test" profile?

optaweb-tsp-planner/pom.xml
yurloc: I agree that commented code is evil most of the times. This is for convenience when exploring various other mutators. So when you're satisfied with your mutation coverage and you wonder what other mutations might survive, you can just uncomment on of those mutators without a) searching for them in Pitest documentation, b) copy-pasting the code.

The reason why they're disabled is that these mutations are not that dangerous and important to cover and so they would just clutter the report with mutations that we don't really care about at this point.

Do you think adding an explanation like this is sufficient or it's better to remove the commented code?

optaweb-tsp-planner/src/main/java/org/optaweb/tsp/optawebtspplanner/TspMapController.java
yurloc: +1 I'll push it directly to master.

optaweb-tsp-planner/pom.xml
rsynek: Let's keep it there if it's convenient for you and add some explanation. I was rather curious what's the reason - although I suspected it's probably there because of  experimenting, it's good to have it written there.

yurloc: My mistake. I checked out the docs and found out that the commented out mutators are part of the `DEFAULT` group so they're enabled now. I have removed them.

yurloc:
@rsynek Please check again.

README.adoc
```diff
@@ -5,3 +5,66 @@ image:https://travis-ci.org/kiegroup/optaweb-vehicle-routing.svg?branch=master[
...
+
+== Running the application
+
+First, clone this repository and `cd` into it.
+
+Frontend and backend are built and run independently. Open two terminal windows.
+
+== Run frontend
+
+. Install `npm`.
```
ge0ffrey: =>
Install npm. If you're on fedora you can run:

README.adoc
```diff
@@ -5,3 +5,66 @@ image:https://travis-ci.org/kiegroup/optaweb-vehicle-routing.svg?branch=master[
...
++
+```shell
+dnf install npm
```
ge0ffrey: =>
sudo dnf install npm

README.adoc
```diff
@@ -5,3 +5,66 @@ image:https://travis-ci.org/kiegroup/optaweb-vehicle-routing.svg?branch=master[
...
+```
+
+. Install dependencies.
++
+```shell
```
ge0ffrey: In the optaplanner docs, we always use this ascciidoc syntax for pieces of code:

```
[source,xml]
----
    <dependency>
      <groupId>org.optaplanner</groupId>
      <artifactId>optaplanner-benchmark</artifactId>
    </dependency>
----
```

We'll probably want to keep that consistent?

yurloc: +1, I didn't realize I'm using Markdown syntax.

README.adoc
```diff
@@ -5,3 +5,72 @@ image:https://travis-ci.org/kiegroup/optaweb-vehicle-routing.svg?branch=master[
...
+JDK 8 is the only system prerequisite. To install OpenJDK 8 on Fedora, run:
+
+[source,shell]
+----
+sudo dnf install java-1.8.0-openjdk-devel
+----
+
+. Download OpenStreetMap data for Belgium from Geofabrik.
+.. Open http://download.geofabrik.de/europe/belgium.html.
+.. Download the PBF file http://download.geofabrik.de/europe/belgium-latest.osm.pbf[belgium-latest.osm.pbf].
```
ge0ffrey: I wouldn't direclty store it under the `local` dir, but under a subdir of it. This subdir should be a directory where we can store all the input pbf files such as belgium, france, europe, us, world.
Those pbf files should not be mixed with other typical local dir files, such as notepad1.txt, benchmarkReportDir, etc.
Also, the graphhopper parsed indexed files that it generated based on the pbf files, should be in a separate subdirectory of local too. (Those should be in local too - so not in target - because we don't want to loose them with "mvn clean").

README.adoc
```diff
@@ -64,7 +64,7 @@ sudo dnf install java-1.8.0-openjdk-devel
 
 . Download OpenStreetMap data for Belgium from Geofabrik.
 .. Open http://download.geofabrik.de/europe/belgium.html.
-.. Download the PBF file http://download.geofabrik.de/europe/belgium-latest.osm.pbf[belgium-latest.osm.pbf].
+.. Download the OSM file http://download.geofabrik.de/europe/belgium-latest.osm.pbf[belgium-latest.osm.pbf].
 .. Save it to `optaweb-vehicle-routing/optaweb-tsp-planner/local/belgium-latest.osm.pbf`.
```
ge0ffrey: Dontt forget to adjust it here too :)

PR #40, opened 2018-12-05
Add tslint-airbnb-config

yurloc:


yurloc:
@danielefiungo Please check again. I accepted your suggestion to relax `variable-name` rule and I also relaxed `function-name` rule to allow pascal case for when we declare a React component as a JS function (e.g. `TripData`).

PR #41, opened 2018-12-11
Osmap gitignore improvements

danielefiungo:
To ease the running of the application I've slightly changed the **planner** .gitignore.

Now the folder ./local and the file ./local/openstreetmap/_placehere.osm.pbf.txt will be committed.

### DIFF
--- a/optaweb-tsp-planner/.gitignore
+++ b/optaweb-tsp-planner/.gitignore
@@ -1,7 +1,7 @@
 !.mvn/wrapper/maven-wrapper.jar
 
 /target
-/local
+
 
 # IntelliJ
 /.idea
@@ -20,3 +20,8 @@
 .settings
 .springBeans
 .sts4-cache
+
+
+# OpenStreetMap data
+local/**/*.osm.pbf
+!local/openstreetmap/_placehere.osm.pbf.txt
--- a/optaweb-tsp-planner/.gitignore
+++ b/optaweb-tsp-planner/.gitignore
@@ -22,6 +22,9 @@
 .sts4-cache
 
 
-# OpenStreetMap data
-local/**/*.osm.pbf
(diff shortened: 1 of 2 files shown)


optaweb-tsp-planner/.gitignore
```diff
@@ -1,7 +1,7 @@
 !.mvn/wrapper/maven-wrapper.jar
 
 /target
-/local
```
yurloc: We should keep this. The `local/` dir should be still completely ignored.

danielefiungo: I've moved and changed /local with local/* otherwise 
```
local/openstreetmap/*
!local/openstreetmap/.gitkeep
```
won't work

danielefiungo: @yurloc find this on git docs:
[gitignore docs](https://git-scm.com/docs/gitignore/2.19.0#_pattern_format)

> An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again. **It is not possible to re-include a file if a parent directory of that file is excluded.** Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined. Put a backslash ("\") in front of the first "!" for patterns that begin with a literal "!", for example, "\!important!.txt".

Make sense?

yurloc:
@danielefiungo I've decided to go for a simpler solution (just a `.gitkeep` file - 86257b7cf0bc39b652b93329a669f12a6a614637).

Firsty, it should do the job of creating the `optaweb-tsp-backend/local/openstreetmap/` directory after cloning the repository and it doesn't require modifying the ignore rules.

Secondly, I think that also the placeholder file is unnecessary as the information is already included in Readme and it's also clearly presented to the user in the exception message in case the OSM file is not found.

Please let me know if it doesn't work for you or open a new PR.

PR #43, opened 2018-12-25
Add pre-push hook that runs frontend tests

yurloc:


yurloc:
For convenience, one can use `git push --no-verify` to bypass pre-push hook.

danielefiungo:
> I don't like the fact that `npm test` is executed even if I'm pushing backend-only changes.
> 
> Anyway, let's merge and give it a try. We can revert husky in case it's harming productivity instead of helping it.


Another problem may rise if you try to commit without initial npm install but in this case probably husky won't be installed and the same will be for git hooks .

Anyway I agree, let's give it a try.

PR #47, opened 2019-01-05
PLANNER-1352: Development Guide

yurloc:
* Initial getting started instructions done in 4fb1e14a214d1e3c0769f5931d198b98664f445f.
* Project structure improved and documented, Javadocs added in #44.
* This PR adds development guide for the backend.

### DIFF
--- a/optaweb-vehicle-routing-backend/README.adoc
+++ b/optaweb-vehicle-routing-backend/README.adoc
@@ -8,6 +8,20 @@
 
 == Project Structure
 
+[literal]
+....
+org.optaweb.vehiclerouting
+├── domain
+├── plugin
+│   ├── persistence
+│   ├── planner
+│   ├── routing
+│   └── websocket
+└── service
+    ├── location
+    └── route
+....
+
 === Root Package
 
 The root package name is identical to the Maven Group ID of the project
@@ -39,7 +53,7 @@ Their implementations are provided by Spring IoC container at runtime.
 
 Contains outer layers such as optimization, persistence, routing and web.
 
-== Dependency rules
+== Dependency Rules
 
 Compile-time dependencies are only allowed in the direction from outer layers
 towards the center.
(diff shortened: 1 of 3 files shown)


optaweb-vehicle-routing-backend/pom.xml
```diff
@@ -72,6 +72,12 @@
       <artifactId>graphhopper-reader-osm</artifactId>
       <version>0.11.0</version>
     </dependency>
+    <!-- Enable Spring Boot Automatic Restart, see Development Guide in Readme to learn how to use it. -->
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-devtools</artifactId>
```
ge0ffrey: Side question:
Does spring-boot-devtools work well for you?
I've seen reports on stackoverflow that drools-core.jar and spring-boot-devtools don't get along. Or was it with org.reflections?

optaweb-vehicle-routing-backend/README.adoc
```diff
@@ -59,3 +73,39 @@ Services are injected into plugins, which moves the burden of their
 construction and dependency resolution to the IoC container.
 Secondly, plugins implement services' output boundary interfaces to handle
 use case results, e.g. persisting changes to database, sending response to web UI.
+
+== Development Guide
+
+=== Automatic Restart
+
+https://docs.spring.io/spring-boot/docs/2.1.1.RELEASE/reference/htmlsingle/#using-boot-devtools-restart[Automatic restart]
```
ge0ffrey: Any way we can replace that "2.1.1.RELEASE" with "latest" or "LATEST" or "latestFinal" and it still works?
Our docs have such a symlink, I would expect theirs to have that too.

optaweb-vehicle-routing-backend/pom.xml
yurloc: For the reference, I have found these:
* https://stackoverflow.com/questions/52606152/optaplanner-got-java-reflection-error-in-spring-boot
* https://stackoverflow.com/questions/35746370/optaplanners-drools-working-memory-is-empty

I haven't noticed any issues so far. We'll see. Thanks for mentioning this.

optaweb-vehicle-routing-backend/README.adoc
yurloc: Yes, they use `current`. I'll change it.

Ideal solution would be keeping the version synchronized with our Spring Boot dependency version and verifying the link doesn't break with version upgrade during CI build.

I have filed #49.

Issue #59, opened 2019-01-22
Group UI components by feature


yurloc:
Currently we have containers + components organization. I want
```
src/
|-- ui/
|   |-- Root.tsx # wrapped by store provider, but not connected itself
|   |-- Map/
|   |   |-- __tests__ # maybe in a separate dir?
|   |   |-- index.ts # maybe, with default export of the connected container
|   |   |-- MapContainer.tsx
|   |   `-- <presentation components>
|   |-- OtherFeature/
|   `-- common/ # maybe reusable components
|-- store/ # or state?
|-- websocketClient/
`-- index.ts # creates store, renders <Root> (wrapped by store provider) into root div

yurloc:
## Examples
* https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Dropdown/ (see index.js)
* https://github.com/skidding/flatris/tree/master/web
* https://github.com/theforeman/foreman/tree/1.21.0-RC3/webpack/assets/javascripts/react_app/components
* **!!!** https://github.com/kubevirt/web-ui-components/blob/master/docs/DevGuide.md

Issue #55, opened 2019-01-22
Improve code organization in store ducks


yurloc:
* `ActionType` enum should probably go to `actions.ts`. It doesn't make much sense to keep it separate from action creators.
* Also note the inconsistency: `types.ts` contains **action types** and **action** interfaces (and state interfaces, which are sort of "domain model" for the frontend), whereas `actions.ts`contains **action creators**.
* Consider using more accurate file names (`actionCreators.ts`, `actionTypes.ts`, `domain.ts`).
* Separate state model (e.g. `ITSPRouteWithSegments` etc.) into a its own file (`domain.ts` or `model.ts`) and maybe pull it up directly under `store/` because it's often used in more than one duck. If it stays in the duck, _probably_ don't export it from index.
* Maybe rename `store/types.ts`. Currently its purpose is to contain common type aliases like `ThunkCommand`.
* Action interfaces should probably be part of the duck's external API (exported from `index.ts`). See `websocket/operations.ts` where we need to dispatch `tsp` and `demo` actions.
* React components should always import from duck's external API, never from its inside. State model is currently being imported from `store/duck/types`. That means state module should be either exported from index or pulled up above ducks to `store/`.

## Confusing imports

`tsp/reducer.ts`:
```js
import { ActionType, ITSPRouteWithSegments, TspAction } from './types';
```
Mixing state model with action types.

`store.ts`:
```js
import { IDemo } from './demo/types';
```
Demo is a state slice. From `types`?? Better: `from './demo/stateModel'` or something like that.

## Final thought
One of the main goals of this cleanup is to make it easy to understand how Redux store (and the whole app) works for people that open the project for the first time or just browse the code on GitHub. It's important that files follow SRP (single responsibility principle) and have descriptive names so that:
* It does exactly what it says on the tin.
* They are consistent (same files in different ducks have identical purpose).

yurloc:
See https://github.com/piotrwitek/react-redux-typescript-guide#redux.

yurloc:
https://www.freecodecamp.org/news/scaling-your-redux-app-with-ducks-6115955638be/
Mainly about ducks but also mentions:
> always separate State Management files from UI files

Issue #53, opened 2019-01-17
Improve TSLint configuration


yurloc:
- [x] ["object-literal-sort-keys": false](https://palantir.github.io/tslint/rules/object-literal-sort-keys) - it's incovenient, often hurts readability
  * additional blank lines may be used to [group object properties](https://palantir.github.io/tslint/rules/object-literal-sort-keys/) together
- [x] ["interface-name": false](https://palantir.github.io/tslint/rules/interface-name) - beautiful [comment](https://github.com/Microsoft/TypeScript-Handbook/issues/121#issuecomment-412810455) summarizing the reasons why not to use the `I` prefix convetion
- [ ] maybe ["jsx-no-lambda": true](https://github.com/palantir/tslint-react)
- [x] increase line width to 120 (TypeScript is more verbose than plain JS)

yurloc:
See also https://github.com/piotrwitek/react-redux-typescript-guide#tslint.

yurloc:
Check out AppFormer JS (https://github.com/kiegroup/appformer-js/blob/react-7.3.x-kiegroup/tslint.common.json).

yurloc:
# No lambda in JSX
Maybe later. React Hooks should avoid this dilemma.
* [React Docs: Handling Events](https://reactjs.org/docs/handling-events.html)
* https://github.com/palantir/tslint-react/issues/96 - long discussion
  * https://medium.freecodecamp.org/why-arrow-functions-and-bind-in-reacts-render-are-problematic-f1c08b060e36
  * "mayb
* [React Binding Patterns: 5 Approaches for Handling `this`](https://medium.freecodecamp.org/react-binding-patterns-5-approaches-for-handling-this-92c651b5af56)
* https://stackoverflow.com/questions/43968779/are-lambda-in-jsx-attributes-an-anti-pattern/43968902#43968902 (no, it's not an anti-pattern)
* https://github.com/facebook/create-react-app/issues/4801#issuecomment-409550702 - :+1: :-1:
* [Best solution for stateless function components](https://stackoverflow.com/a/45581165/1691152)
* [Arrow functions and functions are NOT equivalent](https://stackoverflow.com/a/34361380/1691152)

PR #60, opened 2019-01-22
Redux cleanup

yurloc:
@danielefiungo When you have time, please briefly go through the diff and check if you spot anything you don't like. But more importantly, check out the PR and open it in your IDE. Although it's a lot of changes I hope the result should feel very intuitive. It's still ducks based on https://medium.freecodecamp.org/scaling-your-redux-app-with-ducks-6115955638be.

Notable changes:
- Store broken down into two separate ducks (tsp, websocket) mostly working on separate state slices.
- Isolated code working with SockJS and webstomp-client into the `TspClient` class with nice domain-specific interface (`connect()`, `addLocation()`, `loadDemo()`, ...).
- WebSocket duck is responsible for client connection and reconnection and subscription to route topic.
- TSP duck is responsible for adding/removing locations, loading demo and clearing the whole solution.
- Added `redux-thunk` middleware with `TspClient` as an extra argument. The `TspClient` is then injected by `redux-thunk` to thunks in operations as an extra argument after `dispatch` and `state`.

Looking forward to your feedback. Please note that this is not final solution. There's still room for improvement. I've recorded my ideas in #59 #55 #54 #53.

Lastly, some work is already in progress based on this PR. One is test coverage for the Redux store. Another one is a feature that will block Clear button when demo loading is in progress and it will introduce a separate `demo` duck for that purpose.


optaweb-tsp-ui/src/components/TravelingSalesmanProblem.tsx
```diff
@@ -14,39 +14,47 @@
...
  */
 import * as React from 'react';
+import { connect } from 'react-redux';
+import { IAppState } from '../store/configStore';
+import { tspOperations } from '../store/tsp';
 import { ILatLng, ITSPRouteWithSegments } from '../store/tsp/types';
 import LocationList from './LocationList';
 import TspMap from './TspMap';
 
-export interface ITravelingSalesmanProblemProps {
```
danielefiungo: I think that without the **export** of props interface, the component can't be tested using the interface to pass props

yurloc: True, but there is currently no test for this component, so the export is not needed.

PR #68, opened 2019-02-11
PLANNER-1272: Geocoding proof of concept

yurloc:
## Summary
This is the quickest solution to get basic geosearch feature. I'm using [Leaflet.GeoSearch](https://github.com/smeijer/leaflet-geosearch) React component with built-in `OpenStreetMapProvider` that queries public [Nominatim](https://wiki.openstreetmap.org/wiki/Nominatim) service. It has restrictive usage policy but there are 3rd party providers of the same API and we can set up our own instance.

Leaflet.GeoSearch has other built-in providers for Bind, Esri, Google Maps, and Nominatim-based services LocationIQ and OpenCage. It also has a Leaflet look-and-feel control but we're going to use a custom Patternfly based control (something along the lines of what's part of this PR).

Patternfly doesn't seem to have a generic search control with auto-suggestion but I have emulated it quite easily by combining the [TextInput](http://patternfly-react.surge.sh/patternfly-4/components/textinput) React component and part of [OptionsMenu](https://pf4.patternfly.org/components/OptionsMenu/examples/) by using CSS classes directly from `@patternfly/patternfly`, i.e. bypassing `@patternfly/react-core` (@danielefiungo is that a good idea?).

## Missing features (known limitations)
* Show spinner until the response arrives (currently if the query is "wrong" it's not clear that the response has already arrived but it's empty).
* Show results on the map.
* Results menu should be keyboard-navigable.
* Search input should be focused after a result is selected.
* Clicking outside the result menu should hide it.
* Unify or distinguish seemingly duplicate results (results with different `place_id` but with the same address).

## To run the code locally
1. Make sure you have no uncommitted changes.
2. `git fetch origin pull/68/merge:pr68-geocoding`
3. `git checkout pr68-geocoding`

@danielefiungo Please review the code. What do you think about using CSS classes from `@patternfly/patternfly`?
@ge0ffrey No need to go through the code. Take this as an opportunity to run the application and provide any feedback.
@rsynek QE ping. Tests included.

### DIFF
--- a/optaweb-vehicle-routing-frontend/src/components/SearchBox.test.tsx
+++ b/optaweb-vehicle-routing-frontend/src/components/SearchBox.test.tsx
@@ -43,37 +43,6 @@ describe('Search box', () => {
       searchDelay: 1,
     };
 
-    const results: SearchResult[] = [
-      {
-        bounds: [[1, 2], [3, 4]],
-        label: 'London, ON, Canada',
-        raw: { licence: 'License 1' },
-        x: 101,
-        y: 102,
-      },
-      {
-        bounds: [[1, 2], [3, 4]],
-        label: 'London, OH, USA',
-        raw: { licence: 'License 2' },
-        x: 201,
-        y: 202,
-      },
-      {
-        bounds: [[1, 2], [3, 4]],
-        label: 'London, KY, USA',
-        raw: { licence: 'License 2' },
-        x: 301,
-        y: 302,
-      },
-      {
-        bounds: [[1, 2], [3, 4]],
-        label: 'London, UK',
-        raw: { licence: 'License 1' },
-        x: 401,
-        y: 402,
-      },
-    ];
-
     const searchBox = shallow(<SearchBox {...props} />);
     // @ts-ignore
     OpenStreetMapProvider.mock.instances[0].search.mockImplementation(() => results);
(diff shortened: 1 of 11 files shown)


rsynek:
@yurloc will try to review today/tomorrow

optaweb-vehicle-routing-frontend/src/components/SearchBox.tsx
```diff
@@ -0,0 +1,157 @@
...
+          onChange={this.handleTextInputChange}
+        />
+        {results.length > 0 &&
+        <div className="pf-c-options-menu pf-m-expanded">
+          <ul className="pf-c-options-menu__menu">
+            {results.map((result, index) => (
+              <li key={`result${index}`}>
+                <div className="pf-c-options-menu__menu-item">
+                  {result.address}
+                  <Button
```
rsynek: A soft suggestion - instead of creating the "+" button for each suggested place, you could change the "text" cursor into either "pointer" or "default" (http://www.javascripter.net/faq/stylesc.htm) and accept just a click on the list item.

optaweb-vehicle-routing-frontend/src/components/SearchBox.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+  it('should render text input initially', () => {
+    const props: IProps = {
+      addHandler: jest.fn(),
+      searchDelay: 1,
+    };
+    const searchBox = shallow(<SearchBox {...props} />);
+    expect(toJson(searchBox)).toMatchSnapshot();
+  });
+
+  it('should show results when query is entered', async () => {
```
rsynek: Suggestion - this test declares that "should show results when query is entered", but it focuses also on a negative scenario - empty query shows nothing. Could you consider splitting the test?

rsynek:
I noticed that typing e.g. "Kruis" returns just a single suggestion, while adding one more letter -> "Kruise" suggests more places. Intuitively, I would expect the opposite - making the query more specific should reduce the number of suggested options.

rsynek:
@yurloc would you please advise how to measure the test coverage of the new code?

optaweb-vehicle-routing-frontend/src/components/SearchBox.tsx
yurloc: Good suggestion but it's premature. The widget will appear soon in a completely new screen so I'd rather leave this fine tuning until after the new context is more clear. I also want to be careful with unnecessary customization of Patternfly widgets as that means divergence from what should be there ready to use. I'd A) look for a better widget that's available, or B) make (and help) them add the widgets we need.

yurloc:
> I noticed that typing e.g. "Kruis" returns just a single suggestion, while adding one more letter -> "Kruise" suggests more places. Intuitively, I would expect the opposite - making the query more specific should reduce the number of suggested options.

That is a limitation (feature) of [Nominatim](https://github.com/openstreetmap/Nominatim). I think it simply returns exact matches only, so if there is only 1 exact match it ignores other results that match approximately. I couldn't find detailed description of how matching results are filtered on their [Search API documentation](https://nominatim.org/release-docs/develop/api/Search/).

Note that for example it is [forbidden](https://operations.osmfoundation.org/policies/nominatim/) to implement auto-complete search around the Search API of the public Nominatim server.

yurloc:
> @yurloc would you please advise how to measure the test coverage of the new code?

Run
```
npm run coverage
```
and open the HTML report in `coverage/lcov-report/index.html`.

optaweb-vehicle-routing-frontend/src/components/SearchBox.test.tsx
yurloc: Valid suggestion. I was afraid (lazy) of having to duplicate a lot of code in order to get the component to a state with non-empty results before issuing an empty query to be able to verify that the old results are erased. I take another look and found out it's easy to do.

rsynek:
> > I noticed that typing e.g. "Kruis" returns just a single suggestion, while adding one more letter -> "Kruise" suggests more places. Intuitively, I would expect the opposite - making the query more specific should reduce the number of suggested options.
> 
> That is a limitation (feature) of [Nominatim](https://github.com/openstreetmap/Nominatim). I think it simply returns exact matches only, so if there is only 1 exact match it ignores other results that match approximately. I couldn't find detailed description of how matching results are filtered on their [Search API documentation](https://nominatim.org/release-docs/develop/api/Search/).
> 
> Note that for example it is [forbidden](https://operations.osmfoundation.org/policies/nominatim/) to implement auto-complete search around the Search API of the public Nominatim server.

Interesting, but makes sense.

PR #88, opened 2019-04-06
PLANNER-1299: Add a root (aggregate) POM

yurloc:


optaweb-vehicle-routing-frontend/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+        <artifactId>frontend-maven-plugin</artifactId>
+        <version>1.7.5</version>
+        <executions>
+          <execution>
+            <id>install node and npm</id>
+            <goals>
+              <goal>install-node-and-npm</goal>
+            </goals>
+            <configuration>
+              <nodeVersion>v8.12.0</nodeVersion>
```
rsynek: These versions could be defined in properties to simplify maintenance. It's also likely that in the future the build will be connected to kie-parent, which might define the versions.

optaweb-vehicle-routing-frontend/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+  <version>0.0.1-SNAPSHOT</version>
+  <packaging>pom</packaging>
+
+  <name>OptaWeb Vehicle Routing Frontend</name>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>com.github.eirslett</groupId>
+        <artifactId>frontend-maven-plugin</artifactId>
```
rsynek: I noticed some interesting output messages from this plugin:
warnings https://gist.github.com/rsynek/ff9ffc3c2b4a06a89c1bd017c742099a
warnings about deprecations shown as errors https://gist.github.com/rsynek/a529c7d2616e45d27dd4cd5c0f574a4d
information about successfully running all tests presented as an error: https://gist.github.com/rsynek/b1e5baae33b131d20b4d4baa7ba9ebe1

yurloc: Sure, warnings are true npm warnings that can be observed even when running `npm install` from command line. Unrelated to this PR.

Errors are related to this PR because it's a combination of Jest using `stderr` for output and Maven, displaying `stderr` output as `[ERROR]`. Maven's behavior is more or less correct, at least unsurprising. Jest's behavior is debatable. I'd classify it as a bug as multiple people are [running into it](https://github.com/facebook/jest/issues/5064#issuecomment-401451361). Until they fix it, there's nothing we can do about it.

Deprecation warnings displayed as errors - similar to Jest test report. They're valid warnings unrelated to this PR, but displayed as errors thanks to Maven's translation of `stderr`.

optaweb-vehicle-routing-frontend/pom.xml
yurloc: I don't agree. Using version properties makes maintenance simpler when multiple artifacts share the same version. Or when a value is repeated in multiple plugin configurations. This is not the case.

Ad kie-parent connection. Maybe yes, maybe not. When we start using it we will reuse versions from there. Defining our own properties now won't make the transition easier.

yurloc:
Ready for evaluation. Merging.

PR #89, opened 2019-04-07
PLANNER-1471: Make backend URL configurable during build time

yurloc:


yurloc:
Ready for evaluation, merging into master.

Doesn't allow *build once, deploy multiple" but it should allow to deploy the whole app to OpenShift using S2I build. Backend URL will be picked up from environment during frontend build through `REACT_APP_BACKEND_URL` variable.

PR #90, opened 2019-04-08
Skip Sonar plugin execution on Travis if SONAR_TOKEN is not set (on forks)

yurloc:


optaweb-vehicle-routing-backend/pom.xml
```diff
@@ -361,11 +355,23 @@
...
+    <profile>
+      <!-- Skip Sonar plugin execution if SONARCLOUD_TOKEN is not set. -->
+      <id>skipSonar</id>
+      <activation>
+        <property>
+          <name>!env.SONARCLOUD_TOKEN</name>
+        </property>
+      </activation>
+      <properties>
+        <sonar.skip>true</sonar.skip>
```
yurloc: @rsynek Let me know if you've got an idea how to simplify this. I was thinking about [Travis conditions](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/) but I found this to make more sense in the end.

rsynek: Can you run the analysis manually and pass the token as a property (not an environment variable)? Question nr. 2: do we want to do something like that?

yurloc: Why would I run it manually? I only want to run analysis on PRs and branches in upstream repo from Travis builds (i.e. automatically).

rsynek: You're right, it's probably not needed.

PR #95, opened 2019-04-17
PLANNER-1457,PLANNER-1458: Better support for arbitrary working regions

yurloc:
The UI map component automatically focuses on the loaded OSM file's bounding box. Geosearch results are restricted by the bounding box too and can be further restricted by using one or more country codes in `app.region.country-codes` configuration property.

### DIFF
--- a/optaweb-vehicle-routing-backend/src/main/resources/application.properties
+++ b/optaweb-vehicle-routing-backend/src/main/resources/application.properties
@@ -15,7 +15,7 @@
 #
 app.optimizer.timeout=1m
 app.region.country-codes=BE
-app.routing.osmfile=belgium-latest.osm.pbf
+app.routing.osm-file=belgium-latest.osm.pbf
 
 # Datasource
 # - using an embedded DB with relative path:http://h2database.com/html/features.html#embedded_databases
--- a/optaweb-vehicle-routing-backend/src/main/resources/application.properties
+++ b/optaweb-vehicle-routing-backend/src/main/resources/application.properties
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 app.optimizer.timeout=1m
+app.region.country-codes=BE
 app.routing.osmfile=belgium-latest.osm.pbf
 
 # Datasource
--- a/optaweb-vehicle-routing-backend/src/test/resources/mockito-extensions/README.adoc
+++ b/optaweb-vehicle-routing-backend/src/test/resources/mockito-extensions/README.adoc
@@ -1,0 +1,6 @@
+The Mockito extension file link:org.mockito.plugins.MockMaker[org.mockito.plugins.MockMaker]
+enables
+https://javadoc.io/page/org.mockito/mockito-core/latest/org/mockito/Mockito.html#39[
+Mocking final types].
+
+Needed to mock `com.graphhopper.storage.GraphHopperStorage`.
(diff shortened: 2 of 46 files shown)


optaweb-vehicle-routing-backend/src/main/resources/application.properties
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 app.optimizer.timeout=1m
+app.region.country-codes=BE
 app.routing.osmfile=belgium-latest.osm.pbf
```
ge0ffrey: "country-codes" is plural. So I would expect to be able to fill in multiple values, such as "BE, FR". Is that correct?
If it is, and I do fill in multiple values, can I fill in multiple osm files? I can't, as that parameter is not plural.

We have a convention that multiple values are comma separated.

ge0ffrey: Also, "country-codes" and "osmfile" are not consistent in naming, these are:

- countryCodes and osmFile
- country-codes and osm-file
- countrycodes and osmfile

In any case, we should follow spring boot conventions there.

optaweb-vehicle-routing-backend/src/test/resources/mockito-extensions/README.adoc
```diff
@@ -0,0 +1,6 @@
+The Mockito extension file link:org.mockito.plugins.MockMaker[org.mockito.plugins.MockMaker]
+enables
+https://static.javadoc.io/org.mockito/mockito-core/2.27.0/org/mockito/Mockito.html#39[
+Mocking final types (Mockito 2.27.0 API)].
```
ge0ffrey: Avoid hardcoding versions in docs (or anything outside pom/package files), they tend to go stale easily.
Use this URL instead: https://javadoc.io/page/org.mockito/mockito-core/latest/org/mockito/Mockito.html

optaweb-vehicle-routing-frontend/src/store/websocket/test.ts
```diff
@@ -125,18 +137,75 @@ describe('WebSocket client operations', () => {
...
+
+    // should be subscribed serverInfo topic
+    expect(client.subscribeToServerInfo).toHaveBeenCalledTimes(1);
+
+    store.clearActions();
+
+    // when server info arrives
+    const serverInfo: ServerInfo = {
+      boundingBox: null,
+      countryCodes: ['AB', 'XY'],
```
ge0ffrey: Is it testing with AB and XY?

optaweb-vehicle-routing-backend/src/main/resources/application.properties
yurloc: Both styles (and couple additional ones) work thanks to Spring Boot's [relaxed binding](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-relaxed-binding).

This time I went for `country-codes` because it was suggested by IntelliJ's auto completion. I'll consider that the preferred style and convert all properties mentioned in README and used in `application.properties` to that style.

yurloc: Yes, `app.region.country-codes` is plural and, for example, `app.region.country-codes=DE,AT,CH` is valid.

At this point it's not possible to load multiple OSM files at the same time. `app.routing.osm-file` property is singular and doesn't allow multiple, comma-separated values.

Given the above, it's  possible to load the whole Europe and restrict geosearch to Germany, Austria & Switzerland. I have briefly described this in the readme. I didn't want to go into too much detail as its implementation-dependent and might change soon with adding the possibility to externalize demo data sets.

PR #106, opened 2019-05-20
Model refinement

yurloc:
- Solved the problem with representing a location that doesn't have an ID yet. This is for example a location in an external YAML file. It contains data for creating a new location entity but does not represent a live entity tracked by the system as opposed to a location stored in the database.

  Solved by adding `LocationData(coordinates, description)` and `Location(id) extends LocationData`.

  At one point I tried to avoid the inheritance but in the end it turned out to make sense and to be useful. `RoutingProblem` needs just `LocationData` but it's handy if it accepts `Location` as well.

- Renamed `LatLng` to `Coordinates`. Just because it's easier to talk about it ("coordinates of a location"). Plural because it's latitude and longitude.

- Eliminated getters in domain model.


yurloc:
@rsynek might want to comment on the naming and modeling choices. No need to review the code, it's simple refactoring. The diff may serve as a source of usage examples to see the new names in action.

rsynek:
@yurloc just a single question - what was the reason for getter elimination? I am not saying it's wrong, I just noticed the inconsistency between e.g. `Coordinates` and `PortableCoordinates`.

yurloc:
Yes, that's indeed a small inconsistency. `Portable*` classes use getters so that marshalling with Jackson works with no configuration. It is possible to do `@JsonGetter BigDecimal latitude()` but "portable" API is second-class citizen so the value of being clean and consistent isn't worth the extra code, in my opinion.

For example `PortableLocation` getters have exactly one usage in production code (`WebSocketController` that receives data from client), `PortableCoordinates` getters have zero usages in production code.

`Location` getters (id, coordinates, description) have total of 24 usages.

yurloc:
Some of the reasons for getters elimination are described in https://www.yegor256.com/2014/09/16/getters-and-setters-are-evil.html.

In our simple domain, they're technically still getters but the code is more readable without the "get" prefixes and it encourages a less procedural, more object-oriented way of thinking.

PR #126, opened 2019-06-26
Frontend feedback from employee rostering

yurloc:


optaweb-vehicle-routing-frontend/pom.xml
```diff
@@ -63,6 +63,16 @@
...
+          <execution>
+            <id>npm run lint</id>
+            <goals>
+              <goal>npm</goal>
+            </goals>
+            <phase>test</phase>
+            <configuration>
+              <arguments>run lint</arguments>
+            </configuration>
+          </execution>
```
yurloc: @Christopher-Chianelli you can do the same in `employee-rostering-frontend` to execute `lint` during Maven build.

CC kiegroup/optaweb-employee-rostering#263.

Christopher-Chianelli: I recommend also adding
```
  "husky": {
    "hooks": {
      "pre-push": "npm run lint && CI=true npm test"
    }
  }
```
to `package.json` to ensure linting is done before commiting.

Christopher-Chianelli: Done

yurloc: I'm rather thinking about removing husky. The problem with husky and our use case is that husky works repository-wide (it's a git hook) so it also applies to commits/pushes related to backend. I find it annoying that husky launches npm tests while I'm pushing backend work or even deleting a remote branch with `git push origin :branch-to-delete`.

Even better than linting before commit is to have the linter integrated into IDE and see the errors and warnings on-the-fly. Together with having the habit of auto-formatting everything after finishing code changes and before making a commit, it makes it a very rare event that code style problems are caught no sooner than during CI build (in my workflow).

For completeness, I like to keep the Git history clean so I often do interactive rebases and amend individual commits with typo fixes, code formatting fixes etc.

PR #132, opened 2019-07-19
Prepare the repository to be added to KIE group's repository-list.txt

yurloc:
## Summary of changes
- Root POM is now parent of backend and frontend modules. `kie-parent` is now parent of root POM.
- `spring-boot-starter-parent` is no longer parent of backend but we import `spring-boot-dependencies` for dependency management.
- Overriding a few versions from `kie-parent` in order to build successfully under its dependency management.
- Bound npm scripts to Maven phases that make most sense so if you, for example, run `mvn generate-resources` Maven doesn't run tests and the production build.
- Removed Jacoco and Sonar plugin configurations that are already defined in `kie-parent`. Keeping pull request and branch analysis profiles that activate smartly on Travis.
- Changed SonarCloud project's key, renamed root pom to `org.optaweb.vehiclerouting:optaweb-vehicle-routing`.

## TODO
- [x] Figure out why Pitest takes 5 minutes to execute after the change and fix it (should take around 45s).

## TODO (non-blocking)
- [ ] Review effective pom, override things like description, scm, issues, inception year, etc.
- [ ] Review and unify `maven-compiler-plugin` configuration.
- [ ] Review and unify `pitest-maven` configuration.
- [ ] Avoid version overrides -> upgrade them in kie-parent. Probably not now but in the future. Add TODOs in the code?
- [ ] Remove junit and mockito from kie-parent's dependencies! OK, it still doesn't break anything but now I'm inheriting these in *my* project and I don't like it :smile: (and there's no way to exclude them). And that's a damn good reason to clean it up :smile:.
- [x] Unify profile names for Jacoco and Pitest, remove property activation where it's redundant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ cache:
 install: mvn dependency:go-offline -Pcode-coverage,sonar --show-version
 # do not install to avoid dirtying the cache
 script:
-  - mvn verify -Pcode-coverage,sonar --show-version
+  - mvn verify -Pcode-coverage --show-version
+  - mvn generate-resources -Psonarcloud-analysis
   # check that git working tree is clean after running npm install via a frontend-maven-plugin
   # the git command returns 1 and fails the build if there are any uncommitted changes
   - git diff HEAD --exit-code
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,10 @@ cache:
 # https://docs.travis-ci.com/user/languages/java/#projects-using-maven
 #
 # We also want to avoid Maven's install phase to prevent from writing to ~/.m2/repository, which is cached.
-install: mvn dependency:go-offline -Pjacoco,sonar --show-version
+install: mvn dependency:go-offline -Pcode-coverage,sonar --show-version
 # do not install to avoid dirtying the cache
 script:
-  - mvn verify -Pjacoco,sonar --show-version
+  - mvn verify -Pcode-coverage,sonar --show-version
   # check that git working tree is clean after running npm install via a frontend-maven-plugin
   # the git command returns 1 and fails the build if there are any uncommitted changes
   - git diff HEAD --exit-code
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,26 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-databind</artifactId>
+        <version>2.9.9.1</version>
+      </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.dataformat</groupId>
+        <artifactId>jackson-dataformat-yaml</artifactId>
+        <version>2.9.9</version>
+      </dependency>
+      <dependency>
+        <groupId>com.graphhopper</groupId>
+        <artifactId>graphhopper-reader-osm</artifactId>
+        <version>0.11.0</version>
+      </dependency>
+      <dependency>
+        <groupId>com.neovisionaries</groupId>
+        <artifactId>nv-i18n</artifactId>
+        <version>1.24</version>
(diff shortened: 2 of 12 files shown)


.travis.yml
```diff
@@ -14,10 +14,11 @@ cache:
 # https://docs.travis-ci.com/user/languages/java/#projects-using-maven
 #
 # We also want to avoid Maven's install phase to prevent from writing to ~/.m2/repository, which is cached.
-install: mvn dependency:go-offline -Pjacoco,sonar --show-version
+install: mvn dependency:go-offline -Pcode-coverage,sonar --show-version
```
rsynek: Is it using the existing code-coverage profile from kie-parent? Does it work with SonarCloud well? The profile was added in the past, before any integration with SonarCloud.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/persistence/DistanceRepositoryIntegrationTest.java
```diff
@@ -1,75 +0,0 @@
-/*
```
rsynek: What is the reason for removing this test?

yurloc: Committed by mistake. Thanks for catching it!

.travis.yml
yurloc: Yes, this is `code-coverage` profile from `kie-parent`. It seems to work as expected.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/persistence/DistanceRepositoryIntegrationTest.java
rsynek: There is one more removed test - please see `LocationRepositoryIntegrationTest`

pom.xml
```diff
@@ -19,46 +19,77 @@
...
-  <name>OptaWeb Vehicle Routing Aggregate</name>
+  <name>OptaWeb Vehicle Routing</name>
 
   <modules>
     <module>optaweb-vehicle-routing-backend</module>
     <module>optaweb-vehicle-routing-frontend</module>
   </modules>
 
+  <properties>
+    <jacoco.line.coveredratio.minimum>0.8</jacoco.line.coveredratio.minimum>
```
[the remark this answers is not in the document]
yurloc: Almost no effect. It removes a warning log entry :slightly_smiling_face:. But yes, it's almost no effect. Ideal would be to completely remove the `check` goal.

.travis.yml
yurloc: I tried with a new SonarCloud project after I figured out that I need to change project key. The metrics seemed OK (~80% coverage). I'm pretty confident, we'll get the expected report after merging.

Note: the test inclusion filter seems now to work as expected (tests' coverage is not measured). Overall, sonar plugin behavior changed a lot after inheriting from `kie-parent`. I can show you offline.

rsynek: After quickly looking on the configuration in kie-parent, there is no reason why is should not work for unit tests. Just be aware, that if you migrate the build to jenkins, the profile name triggered in jenkins jobs is different.

pom.xml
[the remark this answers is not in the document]
rsynek: +1. Once we use jacoco together with sonarcloud on every repository, that depends on kie-parent, there is no use to the `check` goal. The same applies to a `spotbugs` maven plugin.

yurloc:
Merging now. I want to have this on master as soon as possible. The next step is to get a green full downstream build (which takes over 6 hours) of kiegroup/droolsjbpm-build-bootstrap#1000.

PR #133, opened 2019-07-23
Exclude a few javax APIs that are already dragged in by Spring Boot

yurloc:
Necessary to comply with Enforcer's `banDuplicateClasses` rule.

This is a temporary solution. Once we drop dependency on `optaplanner-examples`, which we only have to get VRP domain for free, the colliding artifacts will disappear.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,21 @@
         <scope>import</scope>
       </dependency>
       <dependency>
+        <groupId>org.optaplanner</groupId>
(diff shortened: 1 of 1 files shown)


yurloc:
Merging now so that I can execute full downstream build for kiegroup/droolsjbpm-build-bootstrap#1000 ASAP.

CC @rsynek if you have any comments, I'll address them later.

PR #135, opened 2019-07-27
Clean up after adpoting kie-parent

yurloc:


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/websocket/PortableLocation.java
```diff
@@ -85,7 +85,7 @@ public boolean equals(Object o) {
         }
         PortableLocation that = (PortableLocation) o;
         return id == that.id &&
-                Objects.equals(description, that.description) &&
```
rsynek: May I ask what is the reason for removing the `Objects.equals`?

yurloc: Yes. Because I realized `description` can't be null, so I can safely do `descrption.equals(that.description)`.

PR #136, opened 2019-07-27
Re-implement Airbnb ESLint config

yurloc:


optaweb-vehicle-routing-frontend/src/ui/components/Background.tsx
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-/* eslint-disable @typescript-eslint/camelcase */
+/* eslint-disable @typescript-eslint/camelcase,camelcase */
```
Christopher-Chianelli: Are there two rules for the same thing? (I see `@typescript-eslint/camelcase` and `camelcase`, which by spelling appear to be the same rule)

yurloc: I think I had to do this as these are two distinct rules coming from different configs although they both check for the same issue (are equivalent). Let me check if I can eliminate one of them...

yurloc: `@typescript-eslint/camelcase` comes from `plugin:@typescript-eslint/recommended`, `camelcase` comes from `airbnb`. I don't want to remove either of the configs because they don't overlap completely (the typescript config has some useful rules). But since the two rules are equivalent I can disable one of them globally.

PR #143, opened 2019-08-05
Upgrade GraphHopper to fix OSM loading

yurloc:
## Dependency problem and the fix
graphhopper-reader-osm:0.11.0 uses osmosis-osm-binary:0.46, which
requires protobuf-java:3.4.0.

kie-parent sets protobuf version to 3.6.1 but that is not compatible
with osmosis 0.46. Rather than downgrading protobuf, we'll upgrade
graphhopper to latest, which uses osmosis-osm-binary:0.47, which is
compatible with protobuf-java:jar:3.6.1.

## Test
A small custom (random) OSM file from https://download.bbbike.org/osm/ has been added to `/src/test/resources`. It only has 11K. Proper credit is given to https://www.openstreetmap.org/copyright.


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/RoutingConfig.java
```diff
@@ -64,7 +65,7 @@ GraphHopperOSM graphHopper() {
         }
         graphHopper.setOSMFile(osmPath);
         graphHopper.setGraphHopperLocation(GH_DIR + osmPath.replaceFirst(".*/(.*)\\.osm\\.pbf$", "$1"));
-        graphHopper.setEncodingManager(new EncodingManager("car"));
+        graphHopper.setEncodingManager(EncodingManager.create(FlagEncoderFactory.CAR));
```
rsynek: Is seems that the encoding is not backward compatible with an .osm file I have cached locally before this change:

springframework.beans.BeanInstantiationException: Failed to instantiate [com.graphhopper.reader.osm.GraphHopperOSM]: Factory method 'graphHopper' threw exception; nested exception is java.lang.IllegalStateException: Encoding does not match:
Graphhopper config: car|speed_factor=5.0|speed_bits=5|turn_costs=false|version=2
Graph: car|speed_factor=5.0|speed_bits=5|turn_costs=false|version=1
Change configuration to match the graph or delete local/graphhopper/belgium-latest/

But as the error message suggests the solution - deleting the old data - it's not an issue.

yurloc: Yes. Minor rectification: the incompatibility is not related to OSM data but the GraphHopper "graph" data that is a result of parsing OSM and is stored on file system as:
```
$ ls -1 local/graphhopper/belgium-latest
edges
geometry
location_index
names
nodes
nodes_ch_fastest_car
properties
shortcuts_fastest_car
```
So, this graph cache/storage version has changed between graphhopper 0.11 and 0.13. The solution to delete the graph data and let graphhopper regenerate it from OSM is acceptable.

PR #147, opened 2019-08-07
PLANNER-1418: Refactor RouteOptimizerImpl

yurloc:
Follow-up on #144.

`RouteOptimizerImpl` has been broken down into three pieces that now have more limited responsibilities:
- `SolverManager`: runs solver in a different thread, adds problem fact changes, listens for best solution changes.
- `RouteOptimizerImpl`: public API for adding/removing vehicles and locations to solver. Accumulates facts until there's enough data to start solving (solver cannot run with zero entities or empty value ranges).
- `SolutionPublisher`: converts a solution into an event and publishes it through even publisher.

The code is now easier to read and test. This PR also introduces a fix for the corner case when visits are added when there are no vehicles or vice-versa, when there are visits and vehicles are reduced to zero. Creating such condition will now stop solver and prevent it from crashing.

NOTE: Domain model hasn't been updated to accommodate unvisited customers. So even though you can add locations while having zero vehicles, this won't yet reflect in the UI. I'll implement this in the next PR so that we can focus just on the effects of refactoring here.

BONUS: `SolutionUtil` was slowly becoming a big ball of mud and so it was broken down into factories, each responsible for constructing a specific object. Consider guessing a class's purpose from its name:
- `SolutionUtil` - "probably contains some reusable stuff related to solution modification", versus
- `VehicleFactory` - "creates vehicles",
- `SolutionFactory` - "creates solutions" etc.


rsynek:
@yurloc thanks, will take a look tomorrow. Already looking forward to it!

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/SolverManager.java
```diff
@@ -0,0 +1,159 @@
...
+ * <li>Starts solver by running {@link Solver#solve(Object problem)} in a thread that's not the caller's thread.</li>
+ * <li>Stops the solver (synchronously).</li>
+ * <li>Adds problem fact changes to the solver.</li>
+ * <li>Propagates any exception that happens in {@code Solver.solver()} (in a different thread) to the thread that
+ * interacts with {@code SolverManager}.</li>
+ * <li>Listens for best solution changes and publishes new best solutions via {@link SolutionPublisher}.</li>
+ * </ul>
+ */
+@Component
+class SolverManager implements SolverEventListener<VehicleRoutingSolution> {
```
yurloc: @MusaTalluzi FYI. This isn't meant to compete with your SolverManager (I just couldn't come up with a better name). TBH I haven't checked it out for a while so I'm not even sure whether and where they overlap. But definitely has a different goal (see the Javadoc). Please take a look and consider following questions:
- Does this bring something valuable that can be reused in your SolverManager API?
- Does your SM already provide this functionality? Can I drop this when SM is added to optaplanner-core?

Maybe this should be called AsyncSolver.

CC @ge0ffrey.

MusaTalluzi: Thanks Jiri for sharing this, it definitely gives us insights on what should be included in the API. In addition, you having to write this is another indication of the importance of providing SolverManager OOTB so it's less work for users in the future.

For your first question, stopping the solver is something that doesn't exist in the API, I'll add it. All the other features (except for ProblemFactChanges) already exists in SolverManager, I'll let you know before I add it to optaplanner-core to try it and suggest possible changes necessary for VRP.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/SolverManager.java
```diff
@@ -0,0 +1,159 @@
...
+            solver.terminateEarly();
+            // make sure solver has terminated and propagate exceptions
+            try {
+                solverFuture.get();
+                solverFuture = null;
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("Failed to stop solver", e);
+            } catch (ExecutionException e) {
+                throw new RuntimeException("Failed to stop solver", e);
```
rsynek: The `ExecutionException` is just a wrapper, you can use directly the `e.getCause()`.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/SolverManager.java
```diff
@@ -0,0 +1,159 @@
...
+            throw new IllegalStateException("Solver has not started yet");
+        }
+        if (solverFuture.isDone()) {
+            try {
+                solverFuture.get();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("Solver has died", e);
+            } catch (ExecutionException e) {
+                throw new RuntimeException("Solver has died", e);
```
rsynek: ditto

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/RouteOptimizerImplTest.java
```diff
@@ -222,58 +135,145 @@ void removing_wrong_vehicle_should_fail_fast() {
...
+        distanceMap.put(location2.id(), distance);
+        when(distanceMatrix.getRow(location1)).thenReturn(distanceMap);
+        routeOptimizer.addLocation(location1, distanceMatrix);
+
+        verify(distanceMatrix).getRow(location1);
+        VehicleRoutingSolution solution = verifyPublishingPreliminarySolution();
+        assertThat(solution.getDepotList()).hasSize(1);
+        assertThat(solution.getDepotList().get(0).getLocation().getDistanceTo(LocationFactory.fromDomain(location2)))
+                .isEqualTo(8);
+        //FIXME should be: .isEqualTo(distance);
```
rsynek: Do you plan to address this in some next PR?

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/SolverManagerTest.java
```diff
@@ -0,0 +1,218 @@
...
+                .isThrownBy(() -> solverManager.removeLocation(mock(RoadLocation.class)))
+                .withMessageContaining("started");
+    }
+
+    @Test
+    void change_operations_should_fail_is_solver_has_died() throws ExecutionException, InterruptedException {
+        returnSolverFutureWhenSolverIsStarted();
+        solverManager.startSolver(mock(VehicleRoutingSolution.class));
+        when(solverFuture.isDone()).thenReturn(true);
+        when(solverFuture.get()).thenThrow(ExecutionException.class);
```
rsynek: This test is great, still I have an idea how to improve it a bit:
- let's throw an instance of `ExecutionException` (see the overloaded `thenThrow()`) containing the cause
- assert also the cause

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/RouteOptimizerImpl.java
```diff
@@ -16,256 +16,135 @@
...
-                // depot and 1 customer remaining
-                stopSolver();
-                solution.getCustomerList().clear();
-                solution.getLocationList().removeIf(l -> l.getId().equals(location.getId()));
-                solution.getVehicleList().forEach(vehicle -> vehicle.setNextCustomer(null));
-                solution.setScore(HardSoftLongScore.ZERO);
-                publishRoute(solution);
+            if (!visits.removeIf(item -> item.getId().equals(domainLocation.id()))) {
+                throw new IllegalArgumentException(
+                        "Cannot remove " + domainLocation + " because it doesn't exist"
```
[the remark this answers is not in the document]
yurloc: I don't agree with you on this at the moment. I've made a decision not to end exception messages and log messages with a period. For the sake of consistency, I'll keep it this line that way and fix the exception message on line 90. I have reasons for this decision. It's 1) visual aesthetics - similar to why we don't put periods at ends of email subjects or commit message subjects (this is subjective reason to some extent), 2) for consistency with the rest of messages of the whole spring boot app.

I'd like to have a discussion about this and agree on the best approach. I might change my opinion if we find good reasons for the other approach. Source of other points of view: https://stackoverflow.com/questions/1136829/do-you-end-your-exception-messages-with-a-period.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/SolutionFactory.java
```diff
@@ -0,0 +1,130 @@
...
+    private SolutionFactory() {
+        throw new AssertionError("Utility class");
+    }
+
+    /**
+     * Create an empty solution. Empty solution has zero locations, depots, customers and vehicles and a zero score.
+     * @return empty solution
+     */
+    public static VehicleRoutingSolution emptySolution() {
+        VehicleRoutingSolution solution = new VehicleRoutingSolution();
```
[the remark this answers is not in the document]
yurloc: Nope. It has no effect on solving.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/SolutionPublisher.java
```diff
@@ -0,0 +1,138 @@
...
+ * components that listen for this type of event.
+ */
+@Component
+class SolutionPublisher {
+
+    private static final Logger logger = LoggerFactory.getLogger(SolutionPublisher.class);
+
+    private final ApplicationEventPublisher eventPublisher;
+
+    SolutionPublisher(ApplicationEventPublisher eventPublisher) {
```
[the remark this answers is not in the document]
yurloc: Good question. Answer: no. Explanation: https://github.com/spring-projects/spring-framework/blob/v5.2.0.RC1/spring-beans/src/main/java/org/springframework/beans/factory/annotation/AutowiredAnnotationBeanPostProcessor.java#L90-L92.

However this raises another question. How to prevent your question from occurring again? Probably we don't want to add `// @Autowired not needed here` above all bean constructors. I'm starting to think that the best approach is to put `@Autowired` everywhere although it's not required. Yes, I can be smart (lazy) and not use it when it's not needed but that's not friendly towards other people reading the code. Correct?

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/RouteOptimizerImpl.java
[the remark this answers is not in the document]
rsynek: Hmm, there might not be any perfect style of writing error messages. If there is no clear preference, I would say let's use the same convention which is already present in optaplanner-core. And more importantly, let's use any convention consistently across the application.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/SolutionPublisher.java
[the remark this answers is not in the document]
rsynek: Thanks for the link, I was not aware of that. Question is how broad a problem it is. Spring usually follows "convention over configuration".
The comment does not sound like a solution; we either put the annotation there or omit it.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/SolverManagerTest.java
yurloc: This is already covered by `SolverExceptionTest`.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/RouteOptimizerImplTest.java
yurloc: Yes this is related to `optaplanner-examples` where distances are rounded. I'll fix this when forking the planning domain.

PR #149, opened 2019-08-11
PLANNER-1418: Vehicle capacities

yurloc:
- Visits added to `RoutingPlan`. Visits can now be sent to client even though there are no vehicles and no routes (aka "unvisited" customers").
- Vehicles can be added and removed on the Demo page.
- Vehicle capacities are displayed on the Vehicles page and can be increased/decreased by steps of 1.
- Vehicles are not persisted to the database yet. For tech preview this should be OK as we already have persistence for other entities. Will be addressed in a future PR.


rsynek:
Some issues found during local testing:
https://github.com/kiegroup/optaweb-vehicle-routing/issues/150
https://github.com/kiegroup/optaweb-vehicle-routing/issues/151

PR #154, opened 2019-08-20
Remove example dependency

Jaland:
P.R. focused mostly on the removal of the optaplanner example dependencies. Moved the needed code over from the example source, while trying to minimize the amount of code logic I touched. I did remove the need to use the location files from the example code and instead used the native "Location" domain class. 

Extra Notes:
1. Had to remove spring boot dependency cause the class loader was causing issues when the planning solution was moved over
2. Moved the checkstyle logic outside of the pom file.  I think this is better practice as it allows users to sub in there own checkstyle file or remove it if they don't want to use the rules. And if they do want to use it they can integrate the rules more easily into their chosen idea checkstyle plugin. 

TODO: 
1. Update readme (want to make sure the structure looks good first


kie-ci2:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

Jaland:
Looks like I need more test coverage. I will try to get that taken care of at some point this week.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/solver/score/VehicleRoutingEasyScoreCalculator.java
```diff
@@ -0,0 +1,79 @@
...
+import org.optaplanner.core.api.score.buildin.hardsoftlong.HardSoftLongScore;
+import org.optaplanner.core.impl.score.director.easy.EasyScoreCalculator;
+import org.optaweb.vehiclerouting.domain.Customer;
+import org.optaweb.vehiclerouting.domain.Standstill;
+import org.optaweb.vehiclerouting.domain.Vehicle;
+import org.optaweb.vehiclerouting.domain.timewindowed.TimeWindowedCustomer;
+import org.optaweb.vehiclerouting.domain.timewindowed.TimeWindowedVehicleRoutingSolution;
+import org.optaweb.vehiclerouting.solver.VehicleRoutingSolution;
+
+public class VehicleRoutingEasyScoreCalculator implements EasyScoreCalculator<VehicleRoutingSolution> {
```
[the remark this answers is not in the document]
Jaland: Nope, it is commented out in the rule file in the example. Think it was just there to show you that you could do it through java if you wanted.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/solver/VehicleRoutingSolution.java
```diff
@@ -0,0 +1,155 @@
...
+@XStreamAlias("VrpVehicleRoutingSolution")
+@XStreamInclude({TimeWindowedVehicleRoutingSolution.class})
+public class VehicleRoutingSolution extends AbstractPersistable {
+    protected String name;
+    private String distanceUnitOfMeasurement;
+    private List<Location> locationList;
+    private List<Depot> depotList;
+    private List<Vehicle> vehicleList;
+    private List<Customer> customerList;
+    @XStreamConverter(HardSoftLongScoreXStreamConverter.class)
```
yurloc: I see, this is the reason for depending on `optaplanner-persistence`. Not a hard requirement, but you *can* remove all `@XStream` annotations from the migrated code. The planning domain classes are not being serialized in this project.

optaweb-vehicle-routing-backend/pom.xml
```diff
@@ -121,18 +121,18 @@
...
+      <artifactId>optaplanner-persistence-xstream</artifactId>
     </dependency>
     <!-- Include annotation processor that generates metadata for @ConfigurationProperties -->
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-configuration-processor</artifactId>
       <optional>true</optional>
     </dependency>
+    <!-- Had to remove dev tools cause the restart classloader was giving a weird class ClassCastException
+    . https://github.com/AxonFramework/AxonFramework/issues/344-->
```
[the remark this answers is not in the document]
Jaland: Yeah, I think this only caused an issue when I tried to restart through IntelliJ. If I ran it from the command line it did not cause the same issue.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/weight/DepotAngleCustomerDifficultyWeightFactoryTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+
+    /**
+     * Add customer with demand.
+     * @param solution solution
+     * @param location customer's location
+     * @return the new customer
+     */
+    public static PlanningVisit addVisit(VehicleRoutingSolution solution, PlanningLocation location) {
+        PlanningVisit cvisit = new PlanningVisit();
```
yurloc: Typo: cvisit -> visit.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/weight/DepotAngleCustomerDifficultyWeightFactoryTest.java
```diff
@@ -0,0 +1,176 @@
...
+    @Test
+    void createSorterWeight_close_customer_should_have_smaller_weight() {
+        VehicleRoutingSolution solution = createSolution(location1, location2, location3);
+        DepotAngleCustomerDifficultyWeightFactory.DepotAngleCustomerDifficultyWeight closeCustomerWeight =
+                weightFactory.createSorterWeight(solution,
+                solution.getVisitList().get(0));
+        DepotAngleCustomerDifficultyWeightFactory.DepotAngleCustomerDifficultyWeight farCustomerWeight =
+                weightFactory.createSorterWeight(solution,
+                    solution.getVisitList().get(1));
+        assert(closeCustomerWeight.compareTo(farCustomerWeight) < 0);
```
yurloc: The `assert` keyword is for [Java Assertions](https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html). Java Assertions are used to check precondtions, postconditions, invariants, parameter checking, etc. and are *disabled by default*.

In tests we use the `assertThat()` method from [AssertJ](https://assertj.github.io/doc/#overview-what-is-assertj). It has a type-sensitive fluent API and gives nice error descriptions.

optaweb-vehicle-routing-backend/pom.xml
```diff
@@ -229,54 +233,27 @@
         <executions>
           <execution>
             <id>validate</id>
-            <phase>validate</phase>
-            <goals>
-              <goal>check</goal>
-            </goals>
+            <phase/>
```
yurloc: The `<phase>` element cannot be empty. Should be `<phase>none</phase>`.

pom.xml
```diff
@@ -64,18 +64,13 @@
       </dependency>
       <dependency>
         <groupId>org.optaplanner</groupId>
-        <artifactId>optaplanner-examples</artifactId>
+        <artifactId>optaplanner-core</artifactId>
+        <version>${version.org.optaplanner}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.optaplanner</groupId>
+        <artifactId>optaplanner-persistence-xstream</artifactId>
```
yurloc: Please remove this once you get rid of the `HardSoftLongScoreXStreamConverter`.

optaweb-vehicle-routing-backend/src/main/resources/application.properties
```diff
@@ -21,7 +21,7 @@ app.routing.osm-file=belgium-latest.osm.pbf
 
 # Logging
 logging.level.root=info
-logging.level.org.optaplanner.core=info
+logging.level.org.optaplanner.core=debug
```
yurloc: Please revert, the application is not supposed to log on debug level in production. Logging level in local environment can be customized with `application-local.properties`. You can start with an empty file and override any properties you like. The file is ignored so you don't need to worry about having uncommitted changes or committing it unintentionally.

Jaland:
> I think we can merge at this state. There are a couple of minor issues but I can handle them after merging together with other polishing work that needs to be done anyways. @Jaland let me know if you want to address some of them, otherwise I'll merge today.

Probably not going to have time to get to it today, so if you want to go ahead and do the merge sounds good to me.

optaweb-vehicle-routing-backend/pom.xml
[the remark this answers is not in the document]
yurloc: Not related to DROOLS-1540. It's https://stackoverflow.com/questions/52586747/classcastexception-occured-on-solver-solve and the solution is to use context classloader when creating `SolverFactory`.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/domain/PlanningLocation.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * The angle relative to the direction EAST.
+     * @param location never null
+     * @return in Cartesian coordinates
+     */
+    public double getAngle(PlanningLocation location) {
+        // Euclidean distance (Pythagorean theorem) - not correct when the surface is a sphere
+        double latitudeDifference = latitude.subtract(latitude).longValue();
+        double longitudeDifference = longitude.subtract(longitude).longValue();
```
yurloc: This is `x - x` which is always zero. Breaks weight comparison based on angle. CH no longer produces pizza-like solution and finishes with much worse score. Will be fixed in #283.

PR #155, opened 2019-08-21
PLANNER-1418: Vehicle editing

yurloc:
https://issues.jboss.org/browse/PLANNER-1418

Feature completed.

Radek has identified a few issues related to over-constrained planning. Their solution is out of scope of the vehicle editing and will be addressed in the future in https://issues.jboss.org/browse/PLANNER-1585.


rsynek:
Let's just wait for the last check to finish before merging.

PR #163, opened 2019-09-04
BXMSPROD-495 7.26.x distribution not excluded from assembly

Ginxo:
@yurloc I finally had to add distribution folder to the sources folder, otherwise the source code from distribution assembly file won't compile.


mareknovotny:
@Ginxo this PR should very probably point to 7.26.x branch instead of master as there is already #164

Ginxo:
the one for 7.26.x https://github.com/kiegroup/optaweb-vehicle-routing/pull/167

PR #167, opened 2019-09-05
BXMSPROD-495 7.26.x distribution not excluded from assembly

Ginxo:
@yurloc I finally had to add distribution folder to the sources folder, otherwise the source code from distribution assembly file won't compile.


mareknovotny:
it requires rebase @Ginxo

PR #168, opened 2019-09-05
Add sanity check script for the distribution assembly

yurloc:
Add Travis script that will fail the build if distribution isn't assembled correctly (it only runs `mvn validate` in it).


yurloc:
FYI @mareknovotny. Maybe you'd like to add similar check to the Jenkins job (we might abandon Travis one day and lose the check).

PR #175, opened 2019-09-13
Expand Checkstyle configuration, fail build in CI only

yurloc:


yurloc:
I think it doesn't do what you're afraid of. You can still have an empty line between fields but no more than 1 empty line.

PR #191, opened 2019-09-13
Do not clone vehicleList, it's a planning entity collection

yurloc:
Planning entity collections are cloned automatically by solution cloner.

### DIFF
--- a/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/change/ChangeVehicleCapacity.java
+++ b/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/change/ChangeVehicleCapacity.java
@@ -38,8 +38,7 @@ public class ChangeVehicleCapacity implements ProblemFactChange<VehicleRoutingSo
             throw new IllegalStateException("Can't look up a working copy of " + vehicle);
         }
 
-        // Note: Unlike facts and fact collections, which are shared between best solutions and working solutions,
-        // planning entities and collections are cloned during solving, so we don't need to clone vehicleList here.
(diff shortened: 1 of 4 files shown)


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/change/ChangeVehicleCapacity.java
```diff
@@ -38,7 +38,11 @@ public void doChange(ScoreDirector<VehicleRoutingSolution> scoreDirector) {
             throw new IllegalStateException("Can't look up a working copy of " + vehicle);
         }
 
-        // TODO should probably clone the vehicle first as it's a fact and it's shared between solution clones
+        // Note: Unlike facts and fact collections, which are shared between best solutions and working solutions,
+        // planning entities and collections are cloned during solving, so we don't need to clone vehicleList here.
+        // To learn more about problem fact changes, see:
+        // https://docs.jboss.org/optaplanner/release/latest/optaplanner-docs/html_single/#problemFactChangeExample
```
ge0ffrey: I might keep the explanation on a higher level - on the "fact clone vs planning clone vs no clone" level - and short:
// No need to fact clone the workingVehicle because it is a planning entity, so it is already planning cloned
// See https://docs.jboss.org/optaplanner/release/latest/optaplanner-docs/html_single/#problemFactChangeExample

Those docs should explain the race conditions if this doesn't happen - the solver thread vs the rest thread - the working solution vs the best solution - the fact clone vs the planning clone vs no clone.

ge0ffrey: As is it's also very good. My comment above is a very soft suggestion, ok to ignore and merge as is.

PR #202, opened 2019-10-07
Do not deploy build artifacts to Maven repository

yurloc:
When I inherited `kie-parent` and plugged this repository into kiegroup build, I didn't realize that all build artifacts will get deployed to JBoss Nexus. I think it is an unnecessary pollution of the Maven repository as none of the artifacts are intended to be depended on by other projects.

@mareknovotny Can you point me to a project that consumes the distribution ZIP if there is one? If not let's skip deploy plugin completely in this repo.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,6 +99,13 @@
           <artifactId>spring-boot-maven-plugin</artifactId>
           <version>${version.org.springframework.boot}</version>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-deploy-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


mareknovotny:
@Ginxo, @almope could you guys check our assembly if we need to have deployed artifacts into maven repository?

yurloc:
Jenkins retest this.

Ginxo:
@mareknovotny it is required by BAPL-1351 but not used in any assembly

mareknovotny:
So as @Ginxo wrote we should not stop deploying it

yurloc:
Obsoleted by #215.

PR #203, opened 2019-10-10
Make it possible to configure Node.js and npm download URLs

yurloc:


optaweb-vehicle-routing-frontend/pom.xml
```diff
@@ -63,8 +70,10 @@
               <goal>install-node-and-npm</goal>
             </goals>
             <configuration>
-              <nodeVersion>v10.16.3</nodeVersion>
+              <nodeVersion>v11.6.0</nodeVersion>
```
mareknovotny: I think we should standartize on the same versions across all repos, but now I will don't object to this version vs. 8.12.0

Christopher-Chianelli: I agree on standardizing node versions across all repos. We can probably do it via a `pluginManagement` within kie-parent:

```
    <pluginManagement>
      <plugins>
        <plugin>
          <groupId>com.github.eirslett</groupId>
          <artifactId>frontend-maven-plugin</artifactId>
          <version>1.7.5</version>
          <configuration>
             <nodeVersion>v11.6.0</nodeVersion>
             <npmVersion>6.9.0</npmVersion>
          </configuration>
       </plugin>
       <!- ... ->
     </plugins>
  </pluginManagement>
```

cuijulian: @ge0ffrey What qualifies a dependency to have a standardized version in `kie-parent`? Once we agree on the dependency versions why not add the missing standardized versions to the bootstrap build? In any case, I can start sifting through any version differences in VRP and ER.

ge0ffrey: Ideally, all dependency versions should be in the kie-parent, to ensure all projects use the same dependency versions. Having 2 different guava versions for example between module A and B, can make projects that use both fail to actually work (NoSuchMethodError and the like), as well as complicate build setup and testing.

PR #208, opened 2019-10-23
Upgrade to the new API for creating SolverConfig

yurloc:
Depends on kiegroup/optaplanner#589.


optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/SolverIntegrationTest.java
```diff
@@ -103,8 +106,9 @@ void removing_visits_should_not_fail() {
                 singletonList(location(2))
         );
 
-        sf.getSolverConfig().setDaemon(true);
-        Solver<VehicleRoutingSolution> solver = sf.buildSolver();
+        solverConfig.setDaemon(true);
```
ge0ffrey: Why not setDaemon(true) in the setup and directly build the SolverFactory there?
both tests set it to true

PR #211, opened 2019-10-29
Remove Spring Boot dependency management

cuijulian:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1086

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
     <version.com.h2database>1.4.199</version.com.h2database>
     <version.javax.validation>2.0.1.Final</version.javax.validation>
     <version.org.assertj>3.11.1</version.org.assertj>
-    <version.org.mockito>2.23.4</version.org.mockito>
+    <version.org.mockito>3.1.0</version.org.mockito>
     <version.org.junit>5.4.2</version.org.junit>
     <version.org.postgresql>42.2.5</version.org.postgresql>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -47,8 +47,6 @@
     <version.org.mockito>2.23.4</version.org.mockito>
     <version.org.junit>5.4.2</version.org.junit>
     <version.org.postgresql>42.2.5</version.org.postgresql>
-    <version.org.springframework>5.1.7.RELEASE</version.org.springframework>
-    <version.org.springframework.boot>2.1.5.RELEASE</version.org.springframework.boot>
   </properties>
 
   <dependencyManagement>
--- a/optaweb-vehicle-routing-backend/pom.xml
+++ b/optaweb-vehicle-routing-backend/pom.xml
@@ -75,6 +75,10 @@
           <groupId>junit</groupId>
           <artifactId>junit</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
+          <artifactId>hamcrest</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>


pom.xml
```diff
@@ -40,10 +40,8 @@
     <version.com.h2database>1.4.199</version.com.h2database>
     <version.javax.validation>2.0.1.Final</version.javax.validation>
     <version.org.assertj>3.11.1</version.org.assertj>
-    <version.org.mockito>2.23.4</version.org.mockito>
+    <version.org.mockito>3.1.0</version.org.mockito>
```
yurloc: @cuijulian Why this?

cuijulian: The older version of Mockito resulted in test errors, with the same exception as in [this GitHub issue](https://github.com/mockito/mockito/issues/1604). Upgrading to the Mockito version that Spring Boot Test Starter uses (3.1.0) resolves this issue. As stated by one of the Spring Boot developers the problem was probably caused by a mixture of Mockito 2.23.4 with the Spring Boot version.

pom.xml
```diff
@@ -72,13 +70,6 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-      <dependency>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-dependencies</artifactId>
-        <version>${version.org.springframework.boot}</version>
-        <type>pom</type>
-        <scope>import</scope>
```
yurloc: TLDR: I need to investigate this.

Haven't had time to review this thoroughly but this looks suspicious to me. How else could you get the Spring Boot managed versions other than importing their BOM? If you remove this here, are you doing the import somewhere else, like the `kie-parent`? If so, that doesn't look like a good idea either -- we don't want to have KIE versions managed by Spring Boot. Unless of course you place the imports in the right order so that `spring-boot-dependencies` is imported first and its dependency management is overridden by later imports of our (KIE) BOMs.

cuijulian: I am indeed importing the `spring-boot-dependencies` BOM in `kie-parent`. I now see my mistake since the Spring Boot BOM would bring in its own managed dependencies into `kie-parent` and override them.  I see two ways to resolve this:

1. Remove the Spring Boot BOM from `kie-parent` and manually add all the Spring Boot dependencies required by the KIE team into the `<dependencyManagement>` section of the parent.
2. Put `spring-boot-dependencies` BOM first in `kie-parent`.

Are there any other options that you think I'm missing? I can ask someone from the build team which option the team prefers.

yurloc: I think that `spring-boot-dependencies` should not be imported in `kie-parent`. That's undesired coupling. Spring Boot and KIE are totally different platforms. In general, KIE projects should not depend on any Spring Boot versions (@mareknovotny please confirm).

I said in general. This (VRP, ER) is a special case where we're building an example Spring Boot application. If we weren't forced to inherit `kie-parent` to conform to the product and community build standards, everything would be nice and simple (for us, maybe not so for the build team).

So what now? I'm not sure I have the right answer but it still makes best sense to me to inherit `kie-parent` and import `spring-boot-dependencies` here. Do we override some versions managed by `kie-parent` as a side effect? I think so (haven't checked) but that's what Spring Boot is about - they have a BOM of artifacts of certain versions that have been tested to work well together.

cuijulian: This makes sense to me. Though to me this defeats the purpose of having `kie-parent` manage Spring Boot dependencies in the first place, if projects that require Spring Boot just override them in their project POMs.

yurloc: Yes, that's absolutely correct and I don't see why `kie-parent` should manage spring boot dependencies. That doesn't make sense to me.

I know, the goal was most likely that ER and VRP and perhaps other Spring Boot apps in KIE use the same Spring Boot version. Could we achieve this just by defining `<version.org.springframework.boot>X.Y</...>` in `kie-parent` and then just use that version in all projects that import `spring-boot-dependencies`. But how do we make sure the version is used?

What if we put `spring-boot-dependencies` in `kie-parent`'s dependency management? So whenever a project imports it, it doesn't have to specify the version. So unless someone overrides `version.org.springframework.boot`, all will use same Spring Boot version and still have Spring Boot dependencies managed by Spring Boot. How does that sound?

cuijulian: From what I understand, the second option is what my [PR](https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1086) currently does, it puts `spring-boot-dependencies` in `kie-parent`'s dependency management while specifying a version. But doesn't this introduce the problem we discussed above in that Spring Boot overrides the dependency management of many transitive dependencies that are managed by KIE? Maybe I misunderstood your last paragraph, please correct me if I'm wrong.

The first option makes more sense to me, we put the Spring Boot version in `kie-parent` (NOT the dependency or dependency management itself) and use that version for all projects that use `spring-boot-dependencies`. Since the KIE projects will have `kie-parent` as a parent POM they should have access to `${version.org.springframework.boot}` when importing `spring-boot-dependencies` in their respective project POMs.

yurloc: I wasn't very clear. The second option I proposed is _almost_ what your PR does but with an important difference (that I didn't realize when I was writing it). What I propose as the second option is to only define the version of `spring-boot-dependencies` in `kie-parent`'s dependency management **without importing it**.

yurloc: As a result, Spring Boot based projects can import `spring-boot-dependencies` without specifying its version. So unless a project overrides that version, all projects will use the same Spring Boot version.

cuijulian: How would this look? Do we just add `<version.org.springframework>VERSION_HERE</version.org.springframework>` in the `<properties>` section of `kie-parent` and leave it at that? This way Spring Boot based projects can put this in their `<dependencyManagement>`:

```
      <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-dependencies</artifactId>
        <version>${version.org.springframework.boot}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
```

And then put the dependency in the `<dependency>` section. Is this what you meant? It looks reasonable to me.

yurloc: That's the first option.

In the second option I imagine we'd also have this in `kie-parent`'s dependency management:
```xml
      <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-dependencies</artifactId>
        <version>${version.org.springframework.boot}</version>
        <type>pom</type>
      </dependency>
```
not sure if `<type>pom</type>` needs to be there but I omitted `<scope>import</scope>`. Child projects could then import it without specifying the version:
```xml
      <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-dependencies</artifactId>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
```
The advantage of this would be that unless you explicitly override the version, you're using same Spring Boot version as everyone else.

cuijulian: Ahh, I see. Yes this makes sense. In my opinion this second option is more clear to read than the first option, since it tells the reader what dependency is being specified. It's also consistent with the other dependency and version management examples in `kie-parent`. Of course, the build team has the final say. I will update my PR.

PR #218, opened 2019-11-07
BXMSPROD-560 lock-treatment-tool execution moved to defult profile

Ginxo:
BXMSPROD-560

Issue #11, opened 2018-09-17
Use haversine formula to calculate distances


yurloc:
https://en.wikipedia.org/wiki/Haversine_formula

PR #223, opened 2019-11-11
PLANNER-1633: Deploy to OpenShift with binary builds and Docker strategy

yurloc:


optaweb-vehicle-routing-backend/pom.xml
```diff
@@ -55,6 +55,11 @@
       <artifactId>h2</artifactId>
       <scope>runtime</scope>
     </dependency>
+    <dependency>
```
Christopher-Chianelli: Is it possible to connect the backend container to another container (which is the postgres database) provisioned by the ./provision script?

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/routing/AirDistanceRouterTest.java
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    void travel_time_should_be_distance_divided_by_speed() {
+        AirDistanceRouter router = new AirDistanceRouter();
+        Coordinates from = Coordinates.valueOf(0, 0);
+        Coordinates to = Coordinates.valueOf(3, 4); // √(3² + 4²) = 5
+        long travelTimeMillis = router.travelTimeMillis(from, to);
+        assertThat(travelTimeMillis).isEqualTo((long) (5
+                * AirDistanceRouter.KILOMETERS_PER_DEGREE
+                / AirDistanceRouter.TRAVEL_SPEED_KPH
+                * 60 * 60 * 1000));
```
Christopher-Chianelli: where did this constant come from?

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/AirDistanceRouter.java
```diff
@@ -0,0 +1,57 @@
...
+import org.optaweb.vehiclerouting.service.distance.DistanceCalculator;
+import org.optaweb.vehiclerouting.service.region.BoundingBox;
+import org.optaweb.vehiclerouting.service.region.Region;
+import org.optaweb.vehiclerouting.service.route.Router;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConditionalOnProperty(prefix = "app.routing", name = "engine", havingValue = "air")
+public class AirDistanceRouter implements Router, DistanceCalculator, Region {
```
Christopher-Chianelli: Depending if you value accuracy over speed, you might want to use Great-circle distance: https://en.wikipedia.org/wiki/Great-circle_distance; your given formula works well enough, although paths with different longitudes might be inaccurate.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/service/region/BoundingBoxTest.java
```diff
@@ -65,4 +67,19 @@ void should_fail_if_bounding_box_has_zero_dimension() {
...
     }
+
+    @Test
+    void constructor_args_not_null() {
+        assertThatNullPointerException().isThrownBy(() -> new BoundingBox(null, Coordinates.valueOf(1.0, 1.0)));
+        assertThatNullPointerException().isThrownBy(() -> new BoundingBox(Coordinates.valueOf(1.0, 1.0), null));
+    }
+
+    @Test
+    void should_work_with_southwest_and_northeast() {
```
Christopher-Chianelli: What happens if you have a north-west coordinate and a south-east coordinate (ex: (10, -10), (-10, 10) )

optaweb-vehicle-routing-frontend/.env
```diff
@@ -0,0 +1,2 @@
+# https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables
+REACT_APP_BACKEND_URL=.
```
Christopher-Chianelli: I think using an empty string will also work

optaweb-vehicle-routing-frontend/package-lock.json
```diff
@@ -1419,6 +1419,14 @@
             "supports-color": "^5.3.0"
           }
         },
+        "rimraf": {
+          "version": "2.7.1",
+          "resolved": "https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
```
Christopher-Chianelli: redhat proxy url; `locktt` been updated to replace these; see https://github.com/kiegroup/optaweb-employee-rostering/pull/320#issuecomment-552459818 (or just use sed)

provision.sh
```diff
@@ -0,0 +1,179 @@
...
+  exit 0
+}
+
+[[ $1 == "--help" ]] && usage
+
+# Check number of arguments and choose summary or display usage.
+case $# in
+  0)
+    summary="No routing config provided. The backend will start in air routing mode.\n\n\
+WARNING: Air mode doesn't give reliable values. \
```
Christopher-Chianelli: This warning is repeated; maybe save it in a bash variable?

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/routing/AirDistanceRouterTest.java
yurloc: It's number of milliseconds in one hour. It's needed to convert travel time in hours (kilometers / speed in kph) to travel time in milliseconds. I've just noticed the literal value is duplicated between the main and test sources so +1 to isolating to a named constant. Thanks!

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/AirDistanceRouter.java
yurloc: Good idea. This was a quick-and-dirty approximation. I'll replace it with https://en.wikipedia.org/wiki/Haversine_formula in the next PR. If we allow the user to configure the travel speed this might actually become a useful feature.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/service/region/BoundingBoxTest.java
yurloc: That's not allowed and is covered by a negative test that has existed before this PR. I've just added a positive test to kill more mutations.

optaweb-vehicle-routing-frontend/.env
yurloc: Thanks, I'll give it a try.

optaweb-vehicle-routing-frontend/package-lock.json
yurloc: Good catch, thanks! Not sure if lock treatment tool will help me avoid this and #218 hasn't been merged yet. It seems I should really set up the commit hook to check this :slightly_smiling_face:.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/AirDistanceRouter.java
yurloc: #11

README.adoc
```diff
@@ -190,4 +189,64 @@ If the application behaves unexpectedly, review the backend terminal output log.
...
+oc new-project _project_name_
+----
+
+4. Run the script:
++
+[source,shell]
+----
+./provision.sh
+----
+
```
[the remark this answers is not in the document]
yurloc: Not that easy. The URL depends on the project name. But `provision.sh` prints the generated URL at the end so I'll mention this.

optaweb-vehicle-routing-backend/pom.xml
yurloc: Yes of course. On OpenShift, the backend and postgresql containers run in different pods. Backend accesses the database pod through a Service called `postgresql`. The service is a result of `oc new-app postgresql-persistent`. I don't specify a name for the app so the Service will be named `postgresql`.

The service name is then hard-coded and consumed in backend's `application.properties`.

optaweb-vehicle-routing-backend/src/main/resources/application.properties
```diff
@@ -25,16 +27,9 @@ logging.level.org.optaplanner.core=info
 logging.level.org.optaweb.vehiclerouting=info
 
 # Datasource
-# - using an embedded DB with relative path:http://h2database.com/html/features.html#embedded_databases
-# - not closing the DB automatically: http://h2database.com/html/features.html#closing_a_database
-spring.datasource.url=jdbc:h2:file:./local/db/vrp;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE
-spring.datasource.username=sa
-spring.datasource.password=
-spring.datasource.driver-class-name=org.h2.Driver
+spring.datasource.url=jdbc:postgresql://postgresql:5432/${DATABASE_NAME}
```
yurloc: @Christopher-Chianelli Here, `postgresql:5432` is the hard-coded name and port of the OpenShift Service created by `oc new-app postgresql-persistent`.

Christopher-Chianelli: What comment are you replying to? I can't seem to find it and I don't recall commenting on it. Are you referring to the `pom.xml` comment? I don't think you need to declare postgres in `pom.xml` to use it; my worry is it get packaged with the JAR for those who don't use it (but I don't think these JAR's are shaded, so they should be good?)

provision.sh
yurloc: I'm afraid it would hurt readability in this case.

provision.sh
```diff
@@ -0,0 +1,179 @@
...
+  echo
+  echo "      oc set env dc/backend APP_ROUTING_ENGINE=graphhopper"
+  echo
+  echo "  The deployment config change will trigger a new pod with the updated environment, which will replace the old pod."
+  echo
+  echo "Example 3."
+  echo "  $(basename "$0") belgium BE http://download.geofabrik.de/europe/belgium-latest.osm.pbf"
+  echo
+  echo "  Sets up backend to run in GraphHopper routing mode and downloads an OSM file from the provided URL \
+on startup. This provides provides the best out-of-the-box experience. No additional manual steps."
```
rsynek: double "provides".

optaweb-vehicle-routing-backend/src/main/resources/application.properties
yurloc: This is in response to https://github.com/kiegroup/optaweb-vehicle-routing/pull/223#discussion_r344987620.

The postgresql dependency in POM is the JDBC driver. I believe that is needed.

optaweb-vehicle-routing-backend/pom.xml
yurloc: https://dzone.com/articles/configuring-spring-boot-for-postgresql

optaweb-vehicle-routing-backend/src/main/resources/application.properties
Christopher-Chianelli: Ah, I thought the dependency was for the entire posgresql database; I will mark the comment as resolve then

optaweb-vehicle-routing-backend/pom.xml
Christopher-Chianelli: Odd, I cannot mark this comment was resolved; so this comment has to do:

MARK AS RESOLVED BY @Christopher-Chianelli

provision.sh
```diff
@@ -0,0 +1,192 @@
...
+  echo
+  echo
+  echo "Example 2."
+  echo "  $(basename "$0") belgium-latest.osm.pbf BE"
+  echo
+  echo "  Use GraphHopper routing mode and upload OSM file or GraphHopper data manually. \
+The backend will be initially configured to air routing mode. When it starts for the first time, \
+use it to upload Belgium OSM file or GraphHopper data to the claimed Persistent Volume. For example:"
+  echo
+  echo "      oc rsync ~/.vrp/openstreetmap/belgium-latest.osm.pbf backend-1-7k5pl:/deployments/local/openstreetmap/"
```
rsynek: oc rsync ./belgium-latest.osm.pbf backend-1-cjtjf:/deployments/local/openstreetmap/
gives me:
error: path ./belgium-latest.osm.pbf must point to a directory

Moreover, the "backend-1-7k5pl" is a name of a running pod. If we advise users to upload the file like that, we should tell them how to find the pod's name.

provision.sh
```diff
@@ -0,0 +1,192 @@
...
+  echo >&2 "  $(basename "$0") region-name country-code-list [osm-file-download-url]"
+  echo >&2 "Example:"
+  echo >&2 "  $(basename "$0") belgium BE http://download.geofabrik.de/europe/belgium-latest.osm.pbf"
+  exit 1
+}
+
+function usage() {
+  echo "Usage:"
+  echo "  $(basename "$0")"
+  echo "  $(basename "$0") region-name country-code-list [osm-file-download-url]"
```
rsynek: Does it mean we accept multiple country codes? What is the typical use case?

provision.sh
```diff
@@ -0,0 +1,192 @@
...
+function wrong_args() {
+  echo >&2 "Wrong number of arguments! Usage:"
+  echo >&2 "  $(basename "$0")"
+  echo >&2 "  $(basename "$0") region-name country-code-list [osm-file-download-url]"
+  echo >&2 "Example:"
+  echo >&2 "  $(basename "$0") belgium BE http://download.geofabrik.de/europe/belgium-latest.osm.pbf"
+  exit 1
+}
+
+function usage() {
```
rsynek: This usage text is a bit heavy to digest. It offers three ways of getting the osm data to the app:
1) no osm data at all - using air distance. We describe it as the first and the simplest way at the same moment we warn the user about downsides of this approach. I would rather read about this possibility after going through the "better" options.

2) uploading local osm file. Requires at least three additional `oc` calls. It's very easy to make mistakes in the steps and the `oc rsync` command does not work for me (see the other comment below). If we want to offer uploading local osm file, let's do it via a file upload dialog in the app, otherwise the UX will suffer.

Description of the additional required steps should in my opinion rather go into documentation - together with a complete example of how to do it, step by step. As a part of usage of this script, it distracts user from understanding how to use the script alone.

3) this is the preferred way, if I understand it correctly. Why is it the last one?

provision.sh
rsynek: The parameter region-name is used in different ways. In example 2, there is exact name of the osm file "belgium-latest.osm.pbf", while in example 3, there is just "belgium". Do we use it in a different context or are just the provided examples misleading?

PR #225, opened 2019-11-15
BXMSPROD-560 Remove productizedProfile with lock-treatment-tool

yurloc:
@Ginxo I think you intended to do this in #218 but it probably disappeared when you merged master to your PR branch.


Ginxo:
@yurloc thanks

PR #227, opened 2019-11-15
Cypress cleanup

yurloc:


optaweb-vehicle-routing-demo-app/pom.xml
```diff
@@ -15,7 +15,8 @@
   ~ limitations under the License.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
rsynek: That's the IntelliJ vs. VS Code formatting. Sorry for that, need to look into VS Code settings.

optaweb-vehicle-routing-demo-app/src/main/resources/static/env-config.js
```diff
@@ -1,20 +0,0 @@
...
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-window.env = {
-  // Set relative backend path. Frontend will look for backend at the same origin it was served from.
-  REACT_APP_BACKEND_URL: '.',
```
rsynek: Once I removed the file, frontend was not able to find backend's URL.

optaweb-vehicle-routing-demo-app/pom.xml
yurloc: I don't think it's IntelliJ vs. VS Code. I just like it better when each statement is on its own line.

PR #229, opened 2019-11-21
Place all frontend test reports in a single subdirectory

rsynek:
Jenkins jobs expect test results to be stored in
**/target/*-reports/TEST-*.xml

### DIFF
--- a/optaweb-vehicle-routing-frontend/cypress.json
+++ b/optaweb-vehicle-routing-frontend/cypress.json
@@ -3,6 +3,7 @@
     "chromeWebSecurity": false,
     "reporter": "junit",
     "reporterOptions": {
-          "mochaFile": "target/test-reports/TEST-cypress.xml"
+          "mochaFile": "target/test-reports/TEST-cypress.xml",
+          "jenkinsMode": true
     }
 }
--- a/optaweb-vehicle-routing-frontend/cypress.json
+++ b/optaweb-vehicle-routing-frontend/cypress.json
@@ -3,6 +3,6 @@
     "chromeWebSecurity": false,
     "reporter": "junit",
     "reporterOptions": {
-          "mochaFile": "target/TEST-cypress.xml"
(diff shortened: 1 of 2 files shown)


yurloc:
This should be accompanied by an update of Jenkins jobs where test reports are matched using something like `**target//surefire-reports/TEST-*.xml`, right?

rsynek:
@yurloc it should already work like that, see:
https://github.com/kiegroup/kie-jenkins-scripts/blob/master/job-dsls/jobs/pr_jobs.groovy#L277

yurloc:
The test is still not visible in Jenkins test report. Can you try adding `"jenkinsMode": true` to reporter options in `cypress.json`?

See https://github.com/michaelleeallen/mocha-junit-reporter.

Result on my machine:
```xml
...
  <testsuite name="Root Suite.Locations can be added and route is computed" timestamp="2019-11-21T14:54:56" tests="1" failures="0" time="4.012">
    <testcase name="Locations added via clicking on a map are added to a route" time="4.012" classname="Locations can be added and route is computed">
    </testcase>
  </testsuite>
...
```
Not sure if that will work as I would expect the `package.class` scheme on the test case level. Even if that doesn't work I'd keep the jenkinsMode in place and report an issue at https://github.com/michaelleeallen/mocha-junit-reporter/issues.

rsynek:
@yurloc the jenkinsMode made it appear in the list of tests - in the root package. The reporter offers no ability to prefix test name with some artificial package. If we want to the test to show under some package in the Jenkins report, we need to change a name of the Cypress test.

yurloc:
Will new build be triggered every time I make a comment?

yurloc:
Or do I just need to utter Jenkins surrounded by spaces?

yurloc:
For the record, the test was there under the root package even before we activated `jenkinsMode`. So it looks like it's broken and we should report/fix it at https://github.com/michaelleeallen/mocha-junit-reporter.

PR #239, opened 2019-12-02
Remove Neovisionaries dependency management

cuijulian:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1130

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,11 +101,6 @@
         <artifactId>graphhopper-reader-osm</artifactId>
         <version>0.13.0-pre13</version>
       </dependency>
-      <dependency>
-        <groupId>com.neovisionaries</groupId>
-        <artifactId>nv-i18n</artifactId>
-        <version>1.24</version>
-      </dependency>
     </dependencies>
   </dependencyManagement>
 


yurloc:
TBH, not sure if this is a good idea. Do we really want versions of ALL dependnecies in every project to be managed in `kie-parent`?

cuijulian:
I'm going to quote @ge0ffrey here: https://github.com/kiegroup/optaweb-vehicle-routing/pull/203#discussion_r337654088

Of course, it's probably not realistic for ALL dep versions in every kie project to be moved to `kie-parent`, but it is an ideal. At least that's my understanding. As for this PR, I've been working on moving OptaWeb project dependency versions to `kie-parent`, and this is one of them. Perhaps my understanding is naive, I'm happy to discuss this and change my mind.

yurloc:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/tree/7.30.0.Final#requirements-for-dependencies

> Do not override versions in kie-parent's pom.xml unless an exception is granted.
If a newer version of the ip-bom already uses the new version, it's of course fine to do a temporarily overwrite in kie-parent's pom.xml.

mareknovotny:
well, the readme needs some love and fixes, but main strategy is there explained correctly ;)

PR #245, opened 2019-12-11
PLANNER-1744: Do not skip deployment of Maven artifacts

yurloc:


mareknovotny:
tagging here @AntStephenson and @Ginxo  fyi this is the fix for our build change so no backport anymore needed

PR #247, opened 2020-01-07
Generate configuration metadata for app.region.country-codes

yurloc:


yurloc:
Out of date. We have moved from Spring Boot to Quarkus (#445).

PR #255, opened 2020-01-23
Avoid unnecessary plugin management

yurloc:
https://maven.apache.org/pom.html#Plugin_Management

It makes no sense to use `<pluginManagement/>` in a leaf POM. By removing unnecessary plugin management we achieve less code and better coherence.


yurloc:
Jenkins retest this.

PR #260, opened 2020-01-31
PLANNER-1797: Move PostgreSQL to a production profile

yurloc:
The application now starts with file-based H2 by default, again. This
provides smoother getting started experience. To use a production-ready
database, e.g. on OpenShift, you have to explicitly activate Spring
profile called "production" like this:
```
java -jar optaweb*backend*.jar --spring.profiles.active=production
```


yurloc:
Cool! Will merge once we have a green build and replicate this to Employee Rostering after we're done with https://github.com/kiegroup/optaweb-employee-rostering/pull/377/.

PR #265, opened 2020-02-18
PLANNER-1761: Use OptaPlanner Spring Boot Starter

yurloc:


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/OptaWebVehicleRoutingApplication.java
```diff
@@ -16,15 +16,18 @@
...
 import org.springframework.scheduling.annotation.EnableAsync;
 
 @SpringBootApplication
 @EnableAsync
 public class OptaWebVehicleRoutingApplication {
 
     public static void main(String[] args) {
-        SpringApplication.run(OptaWebVehicleRoutingApplication.class, args);
+        new SpringApplicationBuilder(OptaWebVehicleRoutingApplication.class)
+                .properties("optaplanner.solver-config-xml=" + Constants.SOLVER_CONFIG)
```
ge0ffrey: This is ok. Note that you can set it in application.properties too instead, fwiw.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/SolverIntegrationTest.java
```diff
@@ -74,7 +74,7 @@
 
     @BeforeEach
     void setUp() {
-        solverConfig = SolverConfig.createFromXmlResource(RouteOptimizerConfig.SOLVER_CONFIG);
+        solverConfig = SolverConfig.createFromXmlResource(Constants.SOLVER_CONFIG);
```
ge0ffrey: Would it make sense to inject the SolverConfig into this test instead?

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/OptaWebVehicleRoutingApplication.java
ge0ffrey: Ah, because the Solver config constant can be reused then. Yes, that makes sense.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/planner/SolverIntegrationTest.java
yurloc: Probably yes but I would have to turn the test into a `@SpringBootTest` that starts the whole Spring framework. I want to avoid that as much as possible.

PR #266, opened 2020-02-18
PLANNER-1801: Replace score DRL with constraint provider

yurloc:


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/VehicleRoutingConstraintProvider.java
```diff
@@ -0,0 +1,67 @@
...
+                .penalizeLong(
+                        "vehicle capacity exceeded",
+                        HardSoftLongScore.ONE_HARD,
+                        (vehicle, demand) -> demand - vehicle.getCapacity()
+                );
+    }
+
+    private static Constraint distanceFromPreviousStandstill(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PlanningVisit.class)
+                .filter(visit -> visit.getPreviousStandstill() != null)
```
triceo: Personally, I'd replace the filter with `ifExists(..., Joiner.equals(...))`.
Filters are not indexed, leading to a little drop in performance.
More importantly, I consider that easier to understand than the `null` check.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/VehicleRoutingConstraintProvider.java
```diff
@@ -0,0 +1,67 @@
...
+                .penalizeLong(
+                        "distance from previous standstill",
+                        HardSoftLongScore.ONE_SOFT,
+                        PlanningVisit::getDistanceFromPreviousStandstill
+                );
+    }
+
+    private static Constraint distanceFromLastVisitToDepot(ConstraintFactory constraintFactory) {
+        return constraintFactory.from(PlanningVisit.class)
+                .filter(visit -> visit.getNextVisit() == null)
```
triceo: Dtto.

triceo: Same comment, except `ifNotExists(...)`

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/VehicleRoutingConstraintProvider.java
yurloc: Turns out the filter is not necessary here.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/VehicleRoutingConstraintProvider.java
yurloc: Your comment from another line (for conext):
> Personally, I'd replace the filter with ifExists(..., Joiner.equals(...)).
Filters are not indexed, leading to a little drop in performance.
More importantly, I consider that easier to understand than the null check.

I agree that the null check is not nice. I tried to replace it with `ifNotExists()` without success. I'd really like to see your idea of doing that.

On the other hand, what I'd find even more easy to understand is
```java
.filter(PlanningVisit::isLast)
```
I'm inclined to take that approach for the sake of readability even at some performance cost. What do you think?

yurloc: I'm looking for visits that have no next visit. That means "for visit A, not exists visit B, such that B.previousStandstill equals A". So I tried
```java
return constraintFactory.from(PlanningVisit.class)
    .ifNotExists(PlanningVisit.class, equal(PlanningVisit::getPreviousStandstill))
    // then penalize distance to depot
```
but that always gives me a zero penalty.

yurloc: Solved:
```java
return constraintFactory.from(PlanningVisit.class)
    .ifNotExists(PlanningVisit.class, equal(PlanningVisit::getNextVisit, Function.identity()))
```
`Joiners` needs Javadoc.

optaweb-vehicle-routing-backend/pom.xml
```diff
@@ -130,6 +130,11 @@
       <artifactId>spring-boot-configuration-processor</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.optaplanner</groupId>
+      <artifactId>optaplanner-test</artifactId>
```
yurloc, line 135: @ge0ffrey This shouldn't be necessary, I think. The dependency should be provided by `optaplanner-spring-boot-starter`.

ge0ffrey: Can we do that and still have it scoped as a test dependency?

ge0ffrey: Or would we rather just add it to the production dependencies for convenience?

ge0ffrey: @triceo ^^

ge0ffrey: Added to this issue: https://issues.redhat.com/browse/PLANNER-1957

PR #269, opened 2020-02-19
BXMSDOC-5380 Using VRP

emmurphy1:
Details in jira:
[BXMSDOC-5380](https://issues.redhat.com/browse/BXMSDOC-5380)
[Using VRP draft](http://file.ork.redhat.com/~emmurphy/BXMSDOC-5380/)


optaweb-vehicle-routing-docs/src/main/asciidoc/index.html
```diff
@@ -0,0 +1,813 @@
+<!DOCTYPE html>
```
yurloc: HTML files should not be committed.

optaweb-vehicle-routing-docs/src/main/asciidoc/index.adoc
```diff
@@ -1 +1,47 @@
-= OptaWeb Vehicle Routing User Guide
+[id="vrp-quick-start"]
+
+= Using OptaWeb Vehicle Routing
+
+:context: vrp-user-guide
+
+As a developer, you can use OptaWeb Vehicle Routing to optimize your vehicle fleet deliveries.  In this guide, you will create and run a sample OptaWeb Vehicle Routing application.
```
yurloc: I suggest to follow the [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) code style. I hope this doesn't conflict with the docs team code style.

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/proc-download-data.adoc
```diff
@@ -0,0 +1,11 @@
+[id="proc-dowload-data_{context}"]
+
+= Download routing data
+The routing engine requires geographical data to calculate the time it takes vehicles to travel between locations. You must download and store OSM (OpenStreetMap) data files on the local file system before you run OptaWeb Vehicle Routing.
```
yurloc: Missing space between the title and the paragraph. The paragraph is not rendered due to the missing blank line.

optaweb-vehicle-routing-docs/src/main/asciidoc/index.adoc
```diff
@@ -1 +1,47 @@
...
+
+include::modules/con-vrp.adoc[leveloffset=+1]
```
yurloc: Question: What are the prefixes good for? What does `con` mean?

Can we avoid repeating the `vrp` keyword. This whole documentation is about VRP. We would need to use it in every file to be consistent. It doesn't bring and valuable information.

yurloc: I'd prefer if the file name reflected the title of its main section so that it's easy to guess what's inside it. For example this file could be named `con-what-is-vehicle-routing.adoc`.

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/proc-vrp-java.adoc
```diff
@@ -0,0 +1,22 @@
...
+java -version
+----
+. If necessary, install OpenJDK 8.
+* To install OpenJDK 8 on Fedora, enter the following command:
++
+[source,shell]
+----
+sudo dnf install java-1.8.0-openjdk-devel
+----
+COMMENT: For balance, we should include another installation example. What is the most common platform for Optaweb?
```
yurloc: Can you turn this into a `// TODO ...` comment so that we don't commit this accidentally.

Answer: We don't know. The next most common platform could be macOS. I don't think we need to provide Java installation instructions for multiple platforms.

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/proc-vrp-build.adoc
```diff
@@ -0,0 +1,28 @@
+[id="proc-vrp-build_{context}"]
+
+= Building OptaWeb Vehicle Routing
+COMMENT: This seems incomplete or out of context. Flagging to review in the context of the entire guide.
```
yurloc: We don't ship the binary build of the application (optaweb-...-standalone.jar) in the **product** artifacts, just the sources. So users that download RHDM's reference-implementation.zip will only have sources of OptaWeb and need to build it from source. They can't use run.sh.

optaweb-vehicle-routing-docs/src/main/asciidoc/index.adoc
```diff
@@ -1 +1,47 @@
...
+
+== OptaWeb Vehicle Routing quick start
+
+You can get up and running with Optaweb Vehicle Routing in just a few steps.
+
+include::modules/proc-download-data.adoc[leveloffset=+1]
```
yurloc: This and all of the below should be `leveloffset=+2` because quickstart chapter is `h2` so we need to make these sections `h3`.

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/con-vrp.adoc
```diff
@@ -0,0 +1,11 @@
+[id="con-vrp_{context}"]
```
yurloc: Can we use the following custom ID syntax?
```
[#con-vrp_{context}]
```
It's simpler.

optaweb-vehicle-routing-docs/src/main/asciidoc/index.adoc
emmurphy1: You mean in text, rather than in code examples? In actual text, we group sentences in paragraphs. Did I misunderstand?

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/proc-download-data.adoc
emmurphy1: Really? Renders for me. But I don't have a problem adding the line. Done.

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/proc-vrp-java.adoc
emmurphy1: OK

optaweb-vehicle-routing-docs/src/main/asciidoc/index.adoc
emmurphy1: When this renders with CCUTIL, level offset1s are chapter headings. So this should be correct.

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/con-vrp.adoc
emmurphy1: No :-) That syntax is required by the tooling team. Sorry. And we do need the _context placeholder.

optaweb-vehicle-routing-docs/src/main/asciidoc/index.html
emmurphy1: OK I'll remove before I commit.

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/proc-download-data.adoc
emmurphy1: And done for the other modules missing the line as well.

optaweb-vehicle-routing-docs/src/main/asciidoc/index.adoc
emmurphy1: For modular documentation, we use the prefixes so that we as writers can easily identify the type of module the file contains. In addition, the tooling team has recently required us to use the prefixes. The use (or will use) the designation when rendering the docs.

Sure, I can rename the ID, no problem. However, we might change the headings so it makes a little more work to maintain. Also, some of the headings are long. But we can see how it goes.

emmurphy1: I know what you mean about repeating VRP. The reason that we do it is because soon, in the not too distant future, users will be able to find individual modules when they search. For a module, each page is page one. So when you look at a module, it must make sense out of context of the rest of the document. I know its a different way to look at writing. It's a compromise.

optaweb-vehicle-routing-docs/src/main/asciidoc/modules/proc-vrp-build.adoc
emmurphy1: OK. I'll need to work on this later.

optaweb-vehicle-routing-docs/src/main/asciidoc/index.adoc
yurloc: Yes, there seem to be a bit of misunderstanding. Let me clear it up.

I'm talking about plain text (in paragraphs). The rule I referenced requires that each sentence starts on a new line. This rule aims at paragraph with multiple sentences. When following the rule, each Asciidoc source line contains only one sentence. Note that a line break between sentence won't terminate the paragraph so sentences will be still grouped (paragraph is terminated by a *blank line*).

See the link in previous comment for the benefits of this rule.

For example this text (above the comments) has two sentences that are both on line 7. My suggestion is to place a line break between them so the result would look like:
```diff
As a developer, you can use OptaWeb Vehicle Routing to optimize your vehicle fleet deliveries.
In this guide, you will create and run a sample OptaWeb Vehicle Routing application.
```

yurloc:
Superseded by #274

PR #270, opened 2020-02-23
[7.33.x] RHPAM-2734: Minor fixes for downgrade to Spring Boot 2.1.12

cuijulian:
Merge with:

https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1196
https://github.com/kiegroup/optaplanner/pull/701
https://github.com/kiegroup/jbpm-work-items/pull/124
https://github.com/kiegroup/optaweb-employee-rostering/pull/397

### DIFF
--- a/optaweb-vehicle-routing-backend/pom.xml
+++ b/optaweb-vehicle-routing-backend/pom.xml
@@ -47,6 +47,16 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-data-jpa</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.activation</groupId>
+          <artifactId>javax.activation-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>javax.xml.bind</groupId>
+          <artifactId>jaxb-api</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.h2database</groupId>
@@ -104,6 +114,10 @@
         <exclusion>
           <groupId>org.hamcrest</groupId>
           <artifactId>hamcrest</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>javax.xml.bind</groupId>
+          <artifactId>jaxb-api</artifactId>
         </exclusion>
       </exclusions>
     </dependency>


pom.xml
```diff
@@ -106,6 +106,11 @@
         <artifactId>graphhopper-reader-osm</artifactId>
         <version>0.13.0-pre13</version>
       </dependency>
+      <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-api</artifactId>
+        <version>5.6.0</version>
```
mareknovotny: this seems like fixing a previous unnoticed problem with using junit from transitive dependency resolution?

yurloc: Transitive dependencies are intentional when using Spring Boot [Starters](https://docs.spring.io/spring-boot/docs/2.2.4.RELEASE/reference/html/using-spring-boot.html#using-boot-starter).

This dependency has to be added because since [2.2.0.RELEASE](https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-test/2.2.0.RELEASE), `junit-jupiter-api` is part of `spring-boot-starter-test`. Downgrading Spring Boot to 2.1.x made the dependency disappear because `spring-boot-starter-test` [2.1.12.RELEASE](https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-test/2.1.12.RELEASE) doesn't have it.

optaweb-vehicle-routing-backend/pom.xml
```diff
@@ -105,6 +115,10 @@
           <groupId>org.hamcrest</groupId>
           <artifactId>hamcrest</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>javax.xml.bind</groupId>
+          <artifactId>jaxb-api</artifactId>
+        </exclusion>
```
yurloc: This doesn't seem to be necessary.

pom.xml
yurloc: Correction: the comment above is not relevant. `optaweb-vehicle-routing-backend` declares `junit-jupiter-api` dependency so it doesn't use it transitively.

The dependency this PR introduces is a *dependency management* that overrides the version to the latest. This is necessary because one of the tests uses JUnit's `@TempDir` annotation that was introduced in [5.4.0](https://junit.org/junit5/docs/5.4.0/release-notes/).

optaweb-vehicle-routing-backend/pom.xml
yurloc: (Minor suggestion. Doesn't block the PR.)

cuijulian: This exclusion is necessary because there were duplicate classes between this dependency and `jakarta.xml.bind-api`, which broke the build locally from `mvn clean install -Dfull`. Not sure if this is expected behaviour since `jakarta.xml.bind-api` is `runtime` scoped.

pom.xml
mareknovotny: just in case you are aware we use 5.5.2 version in other related kiegroup depMgmt, isn't this confusing to mix more minor versions?

cuijulian: @mareknovotny In that case I can change this dependency management version to 5.5.2. Would that address this concern?

cuijulian:
> Tests are not being run. @rsynek knows more about it.

@yurloc It seems like it's caused by tests using a deprecated method in Mockito:

`verifyZeroInteractions(java.lang.Object...) in org.mockito.Mockito has been deprecated`

According to docs it should be refactored to `verifyNoInteractions(Object... mocks)`. I'll try it locally.

yurloc:
@cuijulian No, the warning has been there for quite some time. I think this is not related.

cuijulian:
@yurloc I think you're right, after refactoring the tests to use the updated Mockito method, the tests are still not being run. However, when I change the `junit-jupiter-api` version to `5.5.2`, the tests run locally! Not sure if it's just my machine. Can you reproduce this?

pom.xml
yurloc: Yes, please `${version.org.junit}` from `kie-parent` instead of hard-coding 5.6.0. In this case I think it makes sense because 5.5.2 is the version used by [spring-boot-starter-test:2.1.12.RELEASE](https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-test/2.1.12.RELEASE).

yurloc:
@cuijulian Confirmed. Probably caused by the version mismatch (just a guess). While at it, can you try removing the exclusions and build locally with `-Dfull`? I didn't get an Enforcer complaint on my box.

yurloc:
@cuijulian Confirmed. Probably caused by the version mismatch (just a guess). While at it, can you try removing the exclusions and build locally with `-Dfull`? I didn't get an Enforcer complaint on my box.

cuijulian:
@yurloc Removing exclusions and doing `-Dfull` build fails Enforcer on my machine. Can confirm this on my personal laptop as well.

cuijulian:
@yurloc Removing exclusions and doing `-Dfull` build fails Enforcer on my machine. Can confirm this on my personal laptop as well.

cuijulian:
@yurloc Removing exclusions and doing `-Dfull` build fails Enforcer on my machine. Can confirm this on my personal laptop as well.

yurloc:
@cuijulian The tests are running on my box with Jupiter API 5.5.2 but on of the tests is failing. Don't know what's going on yet.

cuijulian:
@yurloc Can confirm I get the same thing. The `graphhopper_should_import_and_load_osm_file_successfully` test fails with Jupiter API 5.5.2.

optaweb-vehicle-routing-backend/pom.xml
yurloc: I don't think scope affects this.

https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-test/2.1.12.RELEASE doesn't depend on `jaxb-api` so I'm still convinced this exclusion is redundant (and my local build confirms that).

This a trivial issue. The `7.33.x` branch is effectively dead for this project, so no need for perfection. Feel free to ignore this.

pom.xml
yurloc: Version overrides are tricky. The tests probably don't run because of version mismatch between Jupiter API and the engine. Not sure, but I only managed to run tests with JUnit 5.3.2, which is the version that you get with no overrides.

Please remove this once #273 is merged (it makes the downgrade to 5.3.2 possible by avoiding `@TempDir`, which is 5.4.0+).

optaweb-vehicle-routing-backend/pom.xml
cuijulian: @yurloc I can't confirm this locally, enforcer repeatedly fails without this exclusion :/

To be on the safe side I'll keep this exclusion. But thanks for confirming on your local build.

yurloc: Yes, sure. No problem!

PR #277, opened 2020-03-12
BXMSPROD-592 downstream/upstream pipeline files

Ginxo:


Jenkinsfile
```diff
@@ -0,0 +1,49 @@
...
+                sh 'printenv'
+
+            }
+        }
+        stage('Build upstream projects') {
+            steps {
+                dir("droolsjbpm-build-bootstrap") {
+                    script {
+                        githubscm.checkoutIfExists('droolsjbpm-build-bootstrap', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "$CHANGE_TARGET")
+                        load("upstream.stages")
```
yurloc: Is it possible now or will it be possible in the future to reduce the repositories that will be built before this repository to just those that this project really depends on?

Ginxo: yes, the next step is not make it smarter just building its dependencies (dynamically calculated)

Ginxo:
replaced by https://github.com/kiegroup/optaweb-vehicle-routing/pull/279

PR #283, opened 2020-03-19
Clean up planning domain

yurloc:
Follow-up work related to migrating domain model from `optaplanner-examples` to sources of this project (#154).


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/weight/DepotAngleVisitDifficultyWeightFactory.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * On large data sets, the constructed solution looks like pizza slices.
+ * The order of the slices depends on the {@link PlanningLocation#getAngle} implementation.
+ */
+public class DepotAngleVisitDifficultyWeightFactory
+        implements SelectionSorterWeightFactory<VehicleRoutingSolution, PlanningVisit> {
+
+    @Override
+    public DepotAngleVisitDifficultyWeight createSorterWeight(VehicleRoutingSolution solution, PlanningVisit visit) {
+        PlanningDepot depot = solution.getDepotList().get(0);
```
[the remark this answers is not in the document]
yurloc: Yes. Solver can't run without visits. There cannot be any visits without a depot.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/planner/domain/AbstractPlanningObject.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Exists only to share the ID attribute, which is common to all planning entities and facts. The ID plays two roles:
+ * <ol>
+ * <li>It's a {@link PlanningId} that allows to pull working objects from the score director. This is important for
+ * problem fact changes and potentially for multi-threaded solving.</li>
+ * <li>It's a link to the corresponding domain object.</li>
+ * </ol>
+ */
+public class AbstractPlanningObject {
```
triceo: Why the move away from `AbstractPersistable`? Unless there's a specific technical reason, I'd rather we stay consistent with Optaplanner Examples.

yurloc: Because it's confusing. We don't persist planning domain.

I don't like `extends AbstractPlanningObject` just to get a `@PlanningId` either so I can remove `AbstractPlanningObject` completely.

triceo: Your code, your call. Personally, I don't think this change is well motivated.

yurloc: Why?

triceo: Changing things just because, without a specific technical reason. I agree the current naming may not be ideal, but neither is introducing needless change.

PR #295, opened 2020-04-15
PLANNER-1954: Update quickstart chapater to use runLocally.sh

yurloc:
## Script updates
- Add air mode support to the script.
- Allow to download a region if it has no sub-regions to choose from instead of having to go back.
- Rename VRP dir from `.vrp` to `.optaweb-vehicle-routing`. @ge0ffrey is this OK?

## Documentation updates
- Quickstart chapter uses `runLocally.sh`. It's much simpler now.
- Manual instructions moved to a separate chapter so that they can be used by Windows users.


yurloc:
I'll confirm the name of the config dir (`~/.optaweb-vehicle-routing`) with Geoffrey today and merge after that.

PR #304, opened 2020-05-07
PLANNER-1946: Upgrade to the latest @patternfly/react-core

riccardo-forina:
It looks like a more recent version of Typescript is required, so this
also updates Typescript to the latest version available.

CRA gives a warning using this version of Typescript tho:

```
Creating an optimized production build...
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.5.0

YOUR TYPESCRIPT VERSION: 3.8.3

Please only submit bug reports when using the officially supported version.

=============
```

AFAIK it's benign and can be ignored, but it would be a good idea to
update CRA too.

### DIFF
--- a/optaweb-vehicle-routing-frontend/package-lock.json
+++ b/optaweb-vehicle-routing-frontend/package-lock.json
@@ -1662,36 +1662,24 @@
       "integrity": "sha512-5fA7ZG9AVTNl9lNHfAdQqcBv/PjLp3yfsin1lv0gNSwrP2X4zz2mgFd5WMjeBjenZ5WfNzWQSJTnsqsx/LzMZw=="
     },
     "@patternfly/react-core": {
-      "version": "3.30.1",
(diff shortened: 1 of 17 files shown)


kie-ci10:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

yurloc:
@riccardo-forina Thanks man! The idea of upgrading TypeScript already crossed my mind but I'd have spent several hours investigating other options before trying this.

Can you give this PR another round and update test snapshots so that we have a green build? To do that, you would:
- run `npm test`
- when it finishes hit `U` to update snaphots
- commit the changes and push

Let me know if you don't have the bandwidth to do that. I'll open a new PR and keep your commit to give you proper credit.

yurloc:
ok to test

PR #326, opened 2020-06-05
Remove Toolbar (dead code)

yurloc:


yurloc:
Todo: update the comment and mention that `Toolbar` has been removed and should be replaced with `PageHeaderTools` or remove it completely.

https://www.patternfly.org/v4/documentation/react/overview/upgrade-guide#page

PR #335, opened 2020-07-24
BXMSPROD-883 pull request flow on github actions

Ginxo:
https://issues.redhat.com/browse/BXMSPROD-883

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,3 +1,5 @@
+on: [ workflow_dispatch ]
+
 jobs:
   build-chain:
     runs-on: ubuntu-latest
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,16 @@
+jobs:
+  build-chain:
+    runs-on: ubuntu-latest
+    name: Build Pull Request
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build Chain
+      id: build-chain
+      uses: Ginxo/github-action-build-chain@master
+      with:
+        parent-dependencies: 'optaweb-employee-rostering'
+        child-dependencies: 'kie-wb-distributions'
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,16 @@
...
+  build-chain:
+    runs-on: ubuntu-latest
+    name: Build Pull Request
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build Chain
+      id: build-chain
+      uses: Ginxo/github-action-build-chain@master
+      with:
+        parent-dependencies: 'optaweb-employee-rostering'
```
yurloc, line 11: Why not `parent-dependencies: 'optaplanner'`? This project doesn't depend on optaweb-employee-rostering.

mareknovotny: I agree this dependency definition is not correct, this comes only from the [repository-list.txt](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/script/repository-list.txt). The logical dependency is optaplanner as @yurloc suggested

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,16 @@
+jobs:
+  build-chain:
+    runs-on: ubuntu-latest
+    name: Build Pull Request
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build Chain
+      id: build-chain
+      uses: Ginxo/github-action-build-chain@master
```
mareknovotny, line 9: Could we move this to kiegroup organization?

Ginxo: yes we can, moving it

.github/workflows/pull_request.yml
Ginxo: this is something pending from my side. It also depends on kie-wb-common, for instance. Just trying to reproduce the same behavior we have in Jenkins and once it is stable I will add the proper parent dependencies. Do you agree?

.github/workflows/pull_request.yml
Ginxo: @mareknovotny done!

.github/workflows/pull_request.yml
yurloc: Yes, that makes sense. Let's first reproduce the original process and then optimize it.

PR #342, opened 2020-08-04
PLANNER-2067: Remove location's distance from the matrix

### LINKED ISSUE #83: Clean up distance matrix and database after removing a location

yurloc:
https://issues.redhat.com/browse/PLANNER-2067

Also fixes #83.

### DIFF
--- a/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/GraphHopperRouter.java
+++ b/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/GraphHopperRouter.java
@@ -25,6 +25,7 @@ import com.graphhopper.reader.osm.GraphHopperOSM;
 import com.graphhopper.util.PointList;
 import com.graphhopper.util.shapes.BBox;
 import org.optaweb.vehiclerouting.domain.Coordinates;
+import org.optaweb.vehiclerouting.service.distance.DistanceCalculationException;
 import org.optaweb.vehiclerouting.service.distance.DistanceCalculator;
 import org.optaweb.vehiclerouting.service.region.BoundingBox;
 import org.optaweb.vehiclerouting.service.region.Region;
@@ -72,7 +73,7 @@ class GraphHopperRouter implements Router, DistanceCalculator, Region {
         GHResponse ghResponse = graphHopper.route(ghRequest);
         // TODO return wrapper that can hold both the result and error explanation instead of throwing exception
         if (ghResponse.hasErrors()) {
-            throw new RuntimeException("No route", ghResponse.getErrors().get(0));
+            throw new DistanceCalculationException("No route", ghResponse.getErrors().get(0));
         }
         return ghResponse.getBest().getTime();
     }
--- a/optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/persistence/DistanceRepositoryImplTest.java
+++ b/optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/persistence/DistanceRepositoryImplTest.java
@@ -75,4 +75,16 @@ class DistanceRepositoryImplTest {
                 // which turns -1 into -0, so this test wouldn't kill that mutation without the following:
                 .isNotZero();
     }
+
+    @Test
+    void should_delete_distance_by_location_id() {
+        repository.deleteDistances(from);
+        verify(crudRepository).deleteByFromIdOrToId(from.id());
+    }
+
+    @Test
+    void should_delete_all_distances() {
+        repository.deleteAll();
+        verify(crudRepository).deleteAll();
+    }
 }
--- a/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/persistence/DistanceCrudRepository.java
+++ b/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/persistence/DistanceCrudRepository.java
@@ -19,6 +19,7 @@ package org.optaweb.vehiclerouting.plugin.persistence;
 import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
 import org.springframework.transaction.annotation.Transactional;
 
 /**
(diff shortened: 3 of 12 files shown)


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/GraphHopperRouter.java
```diff
@@ -72,7 +73,7 @@ public long travelTimeMillis(Coordinates from, Coordinates to) {
         GHResponse ghResponse = graphHopper.route(ghRequest);
         // TODO return wrapper that can hold both the result and error explanation instead of throwing exception
         if (ghResponse.hasErrors()) {
-            throw new RuntimeException("No route", ghResponse.getErrors().get(0));
+            throw new DistanceCalculationException("No route", ghResponse.getErrors().get(0));
```
triceo, line 76: Personally, I don't see the need to create a new exception type which doesn't provide anything extra on top of `RuntimeException` - except that it perhaps should've been `IllegalStateException`. A more descriptive message avoids the need for a new exception type.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/persistence/DistanceRepositoryImplTest.java
```diff
@@ -75,4 +75,16 @@ void should_return_negative_number_when_distance_not_found() {
                 // which turns -1 into -0, so this test wouldn't kill that mutation without the following:
                 .isNotZero();
     }
+
+    @Test
+    void should_delete_distance_by_location_id() {
```
triceo, line 80: `asFarAsIKnowWeUseCamelCaseEvenInTestMethodNames`. Although I see that the damage has already been done long before, and pretty consistently.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/GraphHopperRouter.java
yurloc: 1. Generic exceptions shouldn't be thrown. SonarCloud [warns about that](https://sonarcloud.io/project/issues?id=org.optaweb.vehiclerouting%3Aoptaweb-vehicle-routing&issues=AW5u71hIs0lcjfPwd2a8&open=AW5u71hIs0lcjfPwd2a8).
2. Throwing a more specialized exception type makes testing more precise.
3. I'm not against reusing existing Java exceptions but IllegalStateException's [description](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalStateException.html) doesn't fit this situation.

triceo: I agree it shouldn't have been a `RuntimeException`, as that is indeed generic. If not `IllegalStateException` either, then for sure `IllegalArgumentException`, as that's what it is - an illegal combination of arguments.

optaweb-vehicle-routing-backend/src/test/java/org/optaweb/vehiclerouting/plugin/persistence/DistanceRepositoryImplTest.java
yurloc: `theDamageHasBeenDoneALongTimeAgoInAGalaxyFarFarAway`

`im_only_trying_not_to_repeat_the_same_mistake_over_and_over`

On the occasion of starting a project from scratch I took the liberty to try a different approach. I applied it consistently and it has proven to work very well, otherwise I would have reverted it as I have done with other "experiments" that didn't work out very well.

`snake_case` is easier to read and by that it encourages writing a greater amount of small, focused tests that have a clear purpose defined by the method name (which helps to keep the test scope limited). When a test fails it's faster to realize what functionality is broken.

triceo: We seem to have a difference of opinion. I don't understand how having *longer* test names which look *different* than any other names in the code makes anything better.

Still, if test granularity and understanding of results is what you're after, I suggest you explore test nesting, available in JUnit 5.

yurloc: Hmm, nested tests are more about sharing different setup conditions with different groups of test methods. It solves a different problem but yes, it looks useful. I'll try to use it as soon as I come across a test that would benefit from this approach (#344).

BUT, having seen the JUnit 5 docs chapter about nested test classes, I noticed there is another tool that solves exactly the problem `i_tried_to_solve_using_snake_case` test methods names, and that's `@DisplayName("should do something")` (#345).

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/plugin/routing/GraphHopperRouter.java
yurloc: IllegalArgumentException: Thrown to indicate that a method has been passed an illegal or inappropriate argument.

I don't think that applies either. `[100,0]` might be considered an illegal argument as a latitude over 90° is not a valid coordinate.

Anyway, a runtime exception indicates a programming error. As a programmer you have no way to check whether there is a route between `from` and `to` coordinates other than calling the method, so I actually think that a better approach would be to return a response object that holds either a value or an error.

triceo: Yes, some kind of an `Either` construct would be ideal - I was assuming we don't have that.

yurloc: We don't but it's emerging... https://www.baeldung.com/vavr-try.

triceo: I was using that for a while. Not a bad library, but I found that adding a 1MB JAR for 1 class just wasn't worth it for me.

yurloc:
@Christopher-Chianelli can you take a quick look? I'm specifically looking for another pair of eyes for the `CrudRepository` and the new `@Query` and the rest of the magic annotations. You have more experience with Spring Data than me.

PR #349, opened 2020-08-13
PLANNER-2111: Add distance matrix calculation benchmark

yurloc:


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/domain/Location.java
```diff
@@ -61,8 +65,9 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return "Location ["
-                + id +
-                "]: '" + description() + "'";
+        if (description().isEmpty()) {
```
rsynek, line 68: Maybe a ternary operator would make it shorter and more clear?

pom.xml
```diff
@@ -35,6 +35,7 @@
   <!-- Modules are sorted in preferred build order. -->
   <modules>
     <module>optaweb-vehicle-routing-backend</module>
+    <module>optaweb-vehicle-routing-benchmark</module>
```
triceo, line 38: May I suggest we don't call this a "benchmark"?

In the context of OptaPlanner, "benchmark" already has a very specific meaning - the result of the OptaPlanner benchmarker. In fact, Optaweb Employee Rostering does have a module called "benchmark", and it contains exactly this other kind of benchmark. In my opinion, this naming here causes needless confusion.

Also, I don't think that this module needs to exist at all. It's a very specific, one-off thing that might as well have its own repo. And that repo may even be a personal one, not a `kiegroup` one.

optaweb-vehicle-routing-benchmark/src/main/java/org/optaweb/vehiclerouting/benchmark/DataSetGenerator.java
```diff
@@ -0,0 +1,105 @@
...
+        maxDescriptionLength = problem.visits().stream().mapToInt(v -> v.description().length()).max().orElse(0);
+    }
+
+    static LocationData nextDatSetItem(RoutingProblem problem, int i) {
+        return problem.visits().get(i % problem.visits().size());
+    }
+
+    static Coordinates randomize(Coordinates coords) {
+        return Coordinates.valueOf(
+                coords.latitude().doubleValue() + Math.random() * 0.08 - 0.04,
```
rsynek, line 55: Could you please explain the numbers here?

optaweb-vehicle-routing-benchmark/src/main/java/org/optaweb/vehiclerouting/benchmark/StopWatch.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+class StopWatch {
+
+    private final long startTime;
+    private long stopTime;
+    private long lastLap;
+    private int lapCount = 0;
+    private final List<Lap> laps = new ArrayList<>(1010);
```
rsynek, line 31: Why is the ArrayList's capacity initialized with this number?

optaweb-vehicle-routing-benchmark/README.adoc
```diff
@@ -0,0 +1,31 @@
...
+Measures distance matrix calculation for N locations using GraphHopper.
+
+.Prerequisites
+* The Belgium OSM file must exist in `local/openstreetmap/belgium-latest.osm.pbf` relative to the working directory.
+
+== Usage
+
+[source]
+----
+java [THREAD_COUNT] -jar target/optaweb-vehicle-routing-benchmark-VERSION-exec.jar [LOCATION_COUNT] [OUTPUT_FILE_NAME]
```
rsynek, line 12: The [THREAD_COUNT] argument is in my opinion misleading; the reader learns it's a system property in fact just in the middle of the text below, while the other arguments are just simple options.

optaweb-vehicle-routing-benchmark/src/main/java/org/optaweb/vehiclerouting/benchmark/StopWatch.java
yurloc: Because the recommended number of locations to test is 1 - 1000. I originally wanted to measure the calculation time for a 1001th location being added to a matrix with 1000 locations so I made the list size slightly higher.

optaweb-vehicle-routing-benchmark/README.adoc
yurloc: Yes. It's a bit different from the other ones but on conceptual level it's correct. In my opinion, it makes the usage example easier to read and understand.

I considered
```
java [-Djava.util.concurrent.ForkJoinPool.common.parallelism=THREAD_COUNT] \
-jar target/optaweb-vehicle-routing-benchmark-VERSION-exec.jar [LOCATION_COUNT] [OUTPUT_FILE_NAME]
```
too long.

yurloc: Wdyt?

optaweb-vehicle-routing-benchmark/src/main/java/org/optaweb/vehiclerouting/benchmark/StopWatch.java
yurloc: Let's add a comment over there?

optaweb-vehicle-routing-benchmark/src/main/java/org/optaweb/vehiclerouting/benchmark/DataSetGenerator.java
yurloc: That's ± 0.04° (± 4,4 km).

optaweb-vehicle-routing-benchmark/README.adoc
rsynek: I don't think it's too long, but it's very clear regarding in what format is a user supposed to pass the option.

If you want to keep the original notation, let's mention it's this system property right on the first line of the argument's description.

optaweb-vehicle-routing-benchmark/src/main/java/org/optaweb/vehiclerouting/benchmark/DataSetGenerator.java
rsynek: Thanks, maybe a short comment would help.

pom.xml
yurloc: ### Naming

I can rename it to `benchmarks` so that it can accommodate benchmarks for other components in the future. My decision was to make that decision later when the need comes.

In my opinion it's a `benchmark` module no matter if it contains a benchmark for a single component or many components or even different types of benchmarks (micro, macro).

### Home
We're going to share the data next week with a specific group of people. They might want to run the benchmark themselves, so the code should be accessible somewhere.

I'm thinking about trying different GraphHopper modes to see if we can tweak it to get better performance. In that situation, it will come handy. Someone else might get into the same situation so I prefer to have it available in the upstream repo instead of a "random" fork.

We could use a feature branch but the code won't be maintained there so you'd have to rebase on master when you wanted to run the benchmark. The code will be always up to date and ready to run on `master` on the other hand.

### Productization
Everything that lands on master will be eventually shipped with the product. Is there a problem with shipping this? If yes, we can put it on a branch or tweak distribution assembly to exclude this module.

yurloc:
@rsynek Interesting. I didn't observe such variability. Can you try one more time? A possible explanation could be that on the first run GraphHopper builds a graph from the OSM file and that could somehow affect the performance of the following calculation.

pom.xml
yurloc: But yes, I get your point. From a point of view this is an almost one-time thing that will maybe be used by me a couple of times and adds a little value to the project and some maintenance cost. We don't have to merge all of the code that has been written.

If there are strong concerns about adding this to master now, I'll close this and push the code to a feature branch so that it can be shared. We can decide whether it should go to master later.

rsynek:
> @rsynek Interesting. I didn't observe such variability. Can you try one more time? A possible explanation could be that on the first run GraphHopper builds a graph from the OSM file and that could somehow affect the performance of the following calculation.

Consecutive runs are far more similar to each other -> it was truly just the first run.

pom.xml
triceo: To clarify my opinion: I don't see a reason to add this to this repo at all. It's a Maven module - you can keep it as a Maven project in your own Github profile. That way, it can be shared with anyone who wants to use it, and doesn't have any impact on what we ship.

yurloc: Lukas, do you see any good reasons not to add distance matrix benchmark code to this repo? I can feel you're against that idea but I still haven't understood your motivation. What's the risk or cost of landing this on `master`.

triceo: - It is one-off code, with no real product demand driving this. It means maintenance forever, with no one asking for it. It is a self-inflicted burden. I have several useful tools I use for my work in my own profile, and I understand they should not be merged, as they are too specific for a particular purpose. (Case in point: https://github.com/triceo/optaplanner-score-director-factory-benchmark)
- If it were to become an official benchmark, it'd have to be somewhat more scientific and reproducible. As we see in other comments here, this one does not provide consistent results on repeated runs.
- The naming (`benchmark`) is confusing, in the context of the wider OptaPlanner project.

PR #357, opened 2020-08-28
Make DataSetMarshaller null-safe

yurloc:
When a YAML data set contains:

```
vehicles:
```

meaning the attribute is present without a value, the corresponding Java
field is set to null regardless the initial value of the field. We can't
control this. The solution is to make the code working with DataSet
null-safe.

### DIFF
--- a/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/dataset/DataSet.java
+++ b/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/dataset/DataSet.java
@@ -16,7 +16,6 @@
 
 package org.optaweb.vehiclerouting.service.demo.dataset;
 
-import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/dataset/DataSet.java
```diff
@@ -71,7 +70,7 @@ public void setDepot(DataSetLocation depot) {
     /**
      * Visits.
      *
-     * @return visits
+     * @return visits (may be {@code null})
      */
     public List<DataSetLocation> getVisits() {
         return visits;
```
rsynek, line 76: Have you considered returning `Collections.emptyList()` if the list is `null`?

yurloc: Yes. I don't think this class should be responsible for that. I know a nice API should never return `null` but this class is not part of an API. For context, it's used for this:
```java
DataSet unmarshalToDataSet(Reader reader) {
    return mapper.readValue(reader, DataSet.class);
}
```
so it is a dumb DTO and I prefer it to faithfully represent the result of reading a file with the Jackson object mapper. The fact that `DataSet` can contain `null`s is then taken care of in `DataSetMarshaller`.

rsynek (approved):

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/dataset/DataSet.java
yurloc: It is a glue code between my code and the Jackson framework, which [requires](http://fasterxml.github.io/jackson-databind/javadoc/2.11/com/fasterxml/jackson/databind/ObjectMapper.html) the reading/writing class to be a basic POJO so I'd rather avoid doing smart things (like `Optional<String> getName() { ... }`) here.

PR #364, opened 2020-09-10
Move IDE setup instructions

yurloc:


optaweb-vehicle-routing-docs/src/main/asciidoc/appendix-contributing.adoc
```diff
@@ -0,0 +1,33 @@
+[appendix]
+[[contributing]]
+= Contributing
```
emmurphy1, line 3: Contributing to Optaweb Vehicle Routing? Try to make the title more specific.

yurloc:
@Christopher-Chianelli it's probably a good idea to replicate this to Employee Rostering as well to keep the two repos in sync. Any comments?

Christopher-Chianelli (approved):

Christopher-Chianelli:
It would be a good idea to have a CONTRIBUTING page for OptaWeb Employee Rostering; somethings need to be changed though (we don't have mvnw, for example).

yurloc:
@Christopher-Chianelli yes that's correct and you have a valid point. I'm not going to expand the docs in that direction in this PR because that's way outside its scope. I have filed a follow-up issue (#376) to maybe add that piece of information later. On the other hand, the documentation doesn't need to be complete but it needs to be accurate :slightly_smiling_face:.

PR #398, opened 2020-11-10
Clean up 8.x

yurloc:
Cleans up POMs, documentation and Dockerfiles after bumping to OptaPlanner 8.0 and replacing `kie-parent` with `optaplanner-build-parent` in #384.

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaweb-vehicle-routing-backend/pom.xml
+++ b/optaweb-vehicle-routing-backend/pom.xml
@@ -90,13 +90,6 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-test</artifactId>
       <scope>test</scope>
-      <exclusions>
-        <!-- Needed because kie-parent depends on hamcrest-core. -->
-        <exclusion>
-          <groupId>org.hamcrest</groupId>
-          <artifactId>hamcrest</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
@@ -214,9 +207,9 @@
               </dependency>
             </dependencies>
             <!--
-              kie-parent contains Pitest configuration that doesn't work well for this project and it's not possible
-              to change it without a complete override (combine.self="override"), for example excludedClasses,
-              timeoutConstant.
+              optaplanner-build-parent contains Pitest configuration that doesn't work well for this project
+              and it's not possible to change it without a complete override (combine.self="override"),
+              for example excludedClasses, timeoutConstant.
             -->
             <configuration>
               <reportsDirectory>local/pit-reports</reportsDirectory>
--- a/optaweb-vehicle-routing-backend/pom.xml
+++ b/optaweb-vehicle-routing-backend/pom.xml
@@ -141,17 +141,6 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-checkstyle-plugin</artifactId>
-        <executions>
-          <execution>
-            <!-- Skip Checkstyle execution inherited from kie-parent. -->
-            <id>validate</id>
-            <phase>none</phase>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
           <compilerArgs>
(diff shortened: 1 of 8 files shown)


optaweb-vehicle-routing-docs/src/main/asciidoc/quickstart.adoc
```diff
@@ -28,12 +28,12 @@ However, if you have a binary distribution of OptaWeb Vehicle Routing, you only
 java -version
 ----
 
-. If necessary, install OpenJDK 8.
-* To install OpenJDK 8 on Fedora, enter the following command:
+. If necessary, install OpenJDK 11.
+* To install OpenJDK 11 on Fedora, enter the following command:
```
triceo, line 32: If we want to slightly nudge people towards 15, it's `dnf install java-openjdk-latest-devel`.

```suggestion
* To install OpenJDK on Fedora, enter the following command:
```

optaweb-vehicle-routing-docs/src/main/asciidoc/quickstart.adoc
```diff
@@ -28,12 +28,12 @@ However, if you have a binary distribution of OptaWeb Vehicle Routing, you only
...
 +
 [source,shell]
 ----
-sudo dnf install java-1.8.0-openjdk-devel
+sudo dnf install java-11-openjdk-devel
```
triceo, line 36:
```suggestion
sudo dnf install java-latest-openjdk-devel
```

triceo (approved):
LGTM. An optional suggestion.

optaweb-vehicle-routing-docs/src/main/asciidoc/quickstart.adoc
yurloc: Is it safe? Do we guarantee that project always builds and runs with the latest Java? If we had latest (wildcard, not a hardcoded version) Java in the CI matrix this would totally OK.

But can we risk that, say, in a year someone follows this, installs Java 17 and runs into problems? A) We can use it as an early warning about issues with the latest Java if the users reports the issue. B) The user gives up and we lose them.

optaweb-vehicle-routing-backend/pom.xml
```diff
@@ -226,9 +207,9 @@
               </dependency>
             </dependencies>
             <!--
-              kie-parent contains Pitest configuration that doesn't work well for this project and it's not possible
-              to change it without a complete override (combine.self="override"), for example excludedClasses,
-              timeoutConstant.
+              optaplanner-build-parent contains Pitest configuration that doesn't work well for this project
```
rsynek, line 210: I wonder how much we use that Pitest configuration in optaplanner-build-parent. But even if we decide to remove it, it's out of scope of this PR.

rsynek (approved):

optaweb-vehicle-routing-docs/src/main/asciidoc/quickstart.adoc
rsynek: Very good point. Even if we specified "latest" JDK in our CI, we still depend on someone updating the "latest" there (Jenkins, Travis, GHA) soon enough.

We should not use something we don't test.

triceo: If someone uses a SNAPSHOT version, I guess they can expect some problems and we will actually appreciate if they report problems to us.
For a stable version, issues with the latest JDK should already be resolved.

triceo: Plus, I usually update our CI a day or two after the new version of OpenJDK is released.

yurloc: @rsynek If we had literally `latest` Java in CI (which will always be the latest release, as opposed to `15`) then it would be OK in my opinion.

Also, with the 6 month release cadence [1] I can imagine we'll get used to watching the releases and update our CI matrix regularly.

What is the chance of something actually breaking with a new Java release? I think it's highly unpredictable. The code should probably always keep working if you use the right dependencies. The build tools might break. We've had issues when upgrading to Java 11.

[1] https://www.oracle.com/java/technologies/java-se-support-roadmap.html

yurloc:
Taking the safe path now and merging without recommending to install `latest` Java. Let's discuss how to push users to using latest Java separately.

PR #448, opened 2021-03-15
Configure docker to run as non-root user

dupliaka:
The same fix as here:
https://github.com/kiegroup/optaweb-employee-rostering/pull/587


Christopher-Chianelli (approved):

yurloc:
Maven Build (11) didn't start and cannot be started manually. Jenkins linux build has passed, that's enough. Merging.

PR #468, opened 2021-04-07
Add readme and lincese to distribution root

yurloc:
### Referenced pull requests

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaweb-vehicle-routing-distribution/src/main/assembly/assembly-optaweb-vehicle-routing.xml
+++ b/optaweb-vehicle-routing-distribution/src/main/assembly/assembly-optaweb-vehicle-routing.xml
@@ -27,6 +27,14 @@
 
   <files>
     <file>
+      <source>../LICENSE-Apache-2.0.txt</source>
+      <outputDirectory/>
+    </file>
+    <file>
+      <source>src/main/resources/README.adoc</source>
+      <outputDirectory/>
+    </file>
+    <file>
       <source>../runLocally.sh</source>
       <outputDirectory>bin</outputDirectory>
       <filtered>true</filtered>


Christopher-Chianelli (approved):
Side note: should the assembly also produce a directory (which make it easier to check its contents in a review; Employee Rostering produces a directory for this reason)

yurloc:
> Side note: should the assembly also produce a directory (which make it easier to check its contents in a review; Employee Rostering produces a directory for this reason)

Personally, I don't mind doing an `unzip` first during a review, and so, I consider the extra `dir` output **redundant**. I wouldn't really mind a single extra line in the assembly XML. What I do mind is the **warning** I can see when the `dir` output is enabled:

> [WARNING] Assembly file: /home/jlocker/src/github.com/kiegroup/optaweb-employee-rostering/optaweb-employee-rostering-distribution/target/optaweb-employee-rostering-distribution-8.6.0-SNAPSHOT is not a regular file (it may be a directory). It cannot be attached to the project build for installation or deployment.

### Warning noise

My approach in general is to eliminate every single warning and error message if the effort is reasonable because:
1. If it's a good warning that points to a real problem.
  a. You fix the problem and get rid of the warning. All is good.
  b. You ignore it => you have an untreated problem in your code AND a warning increasing the overall warning noise.
2. If it's a false positive.
  a. You eliminate it either by legitimately silencing the warning (for example marking it as a false positive on Sonar Cloud) or by finding a workaround. The cost of the workaround should be lower than the cost of having a false positive warning.
  b. You ignore it => you increase the overall warning noise.

High warning noise comes with a cost.
- It leaks cognitive resources of people looking at the log when investigating a real problem with the build.
- It becomes more difficult to spot new warnings. If you miss a new genuine warning, you miss the opportunity to fix a problem before it's even committed.

### Outcome

Given the benefit I see in reducing warning noise and the low cost of working around this one (removing the `<format>dir</...>` element and having to do `unzip` occasionally) I prefer leaving the `dir` output out.

PR #475, opened 2021-04-21
Do not fail if the graph directory is empty

yurloc:
Improve getting started experience.

People sometimes run into a situation where the OSM import fails (`.osm.pbf` file too large and not enough JVM memory to import it). In that case `~/.optaweb-vehicle-routing/graphhopper/<region>/` is created but it remains empty.

Previously, we only checked whether the directory existed in which case we tried to _load_ the graph and GraphHopper would fail.

Now we also check if the graph dir is empty, in which case we take the _import_ branch.

<details>
<summary>
</summary>

</details>


Ginxo:
Jenkins retest this

rsynek (approved):

PR #511, opened 2021-06-15
Streamify collecting of problems

yurloc:
Just an occasional code cleanup.

<details>
<summary>
</summary>

</details>

### DIFF
--- a/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/RoutingProblemConfig.java
+++ b/optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/RoutingProblemConfig.java
@@ -28,8 +28,6 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/RoutingProblemConfig.java
```diff
@@ -62,38 +60,19 @@
 
     @Produces
     RoutingProblemList routingProblems() {
-        ArrayList<RoutingProblem> problems = new ArrayList<>();
-        problems.add(dataSetMarshaller.unmarshal(belgiumReader()));
-        problems.addAll(localDataSets());
-        return new RoutingProblemList(problems);
+        Stream<RoutingProblem> allProblems = Stream.concat(classPathProblems(), dataSetDirProblems());
```
yurloc, line 63: This is the important change. Instead of building a new array list, simply join two streams.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/RoutingProblemList.java
```diff
@@ -33,8 +33,8 @@
 
     private final Map<String, RoutingProblem> routingProblems;
 
-    RoutingProblemList(List<RoutingProblem> routingProblems) {
-        this.routingProblems = Objects.requireNonNull(routingProblems).stream()
+    RoutingProblemList(Stream<RoutingProblem> routingProblems) {
+        this.routingProblems = Objects.requireNonNull(routingProblems)
```
yurloc, line 37: Stream is required here anyway.

yurloc:
@triceo stream expert review request.

triceo (approved):
The general idea to avoid creating the collections until necessary, I like.
I don't see the general point of the `RoutingProblemList` abstraction.
If you want to keep it, I won't object.

optaweb-vehicle-routing-backend/src/main/java/org/optaweb/vehiclerouting/service/demo/RoutingProblemList.java
```diff
@@ -33,8 +33,8 @@
...
                 // TODO use file name as the key (that's more likely to be unique than data set name)
                 .collect(toMap(RoutingProblem::name, identity()));
```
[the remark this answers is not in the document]
yurloc: I'm also not a fan but what this constructor does is essentially parameter validation and making a defensive copy.

I think it is the responsibility of this class to create the map because the way the map is being built is essential for what this class does. There is the `RoutingProblem byName(String name)` method, raising an expectation that this class is able to fetch problems by their name. By building the index here, inside the class (no matter if it's a constructor or a factory method) it's easy to see and test that the class does what it's supposed to. If I took a map from the outside, I'd have to iterate over it and check all entries.

PR #707, opened 2022-06-07
Cleanup useless Jenkinsfile

radtriste:
File became useless since https://github.com/kiegroup/kogito-pipelines/pull/454

Related PRs:
- https://github.com/kiegroup/optaplanner/pull/2022
- https://github.com/kiegroup/optaweb-vehicle-routing/pull/707
- https://github.com/kiegroup/optaweb-employee-rostering/pull/825
- https://github.com/kiegroup/optaplanner-quickstarts/pull/351


triceo (approved):

radtriste:
@yurloc `(build) optaweb-vehicle-routing` will not be executed because there is no code change.

yurloc:
@radtriste makes sense. Should I replace `(build) optaweb-vehicle-routing` from the required checks? Will there be a replacement for that? I imagine the same Jenkins job, just generated elsewhere, perhaps with a different name?

radtriste:
no replacement, the job still exists. It just does not launch on all PRs. It depends on the content which is updated

yurloc:
> well the Jenkins check is launched only if code is modified. It is not launched when a modification happens only in `.ci` because it makes no sense

I get it now. I originally thought the reason for it not being triggered is the _removal_ of the Jenkins file. The real reason is CI is configured to ignore changes under that path. Thanks for explanation!

yurloc (approved):

PR #798, opened 2022-11-23
[BXMSPROD-1881] dependabot for main, 7.67.x and blue branches

Ginxo:


mareknovotny (approved):

mareknovotny:
i think here 7.67.x and 7.67.x-blue branches doesn't make sense as they are not included. Please remove them and keep only main

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+  target-branch: "main"
+  commit-message:
+    prefix: "[bot][main]"
+- package-ecosystem: maven
+  directory: "/"
+  schedule:
+    interval: daily
+    time: '03:00'
+  open-pull-requests-limit: 10
+  target-branch: "7.x"
```
mareknovotny, line 18: we don't need any optaweb repos for 7.x

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+  commit-message:
+    prefix: "[bot][7.x]"
+- package-ecosystem: maven
+  directory: "/"
+  schedule:
+    interval: daily
+    time: '04:00'
+  open-pull-requests-limit: 10
+  target-branch: "7.67.x"
```
mareknovotny, line 27: we don't need any optaweb repos for 7.x

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+  commit-message:
+    prefix: "[bot][7.67.x]"
+- package-ecosystem: maven
+  directory: "/"
+  schedule:
+    interval: daily
+    time: '05:00'
+  open-pull-requests-limit: 10
+  target-branch: "7.67.x-blue"
```
mareknovotny, line 36: there is no optaplanner and optaweb in blue branches

mareknovotny (changes requested):
we need to remove 7.67.x, 7.x and 7.67.x-blue branches and added 8.13.x to be scanned

mareknovotny (approved):

PR #871, opened 2023-11-13
RHPAM-4700: Upgrade Jackson to 2.15.2 to make it work with SnakeYAML …

mareknovotny:
…2.0 (#870)

https://issues.redhat.com/browse/RHPAM-4700

No org.yaml:snakeyaml:jar:1.33.0 traces on optaweb* or kogito-examples projects after related PRs

Related PRs:
- https://github.com/kiegroup/kogito-runtimes/pull/9
- https://github.com/kiegroup/optaplanner/pull/25

### DIFF
--- a/.ci/jenkins/dsl/jobs.groovy
+++ b/.ci/jenkins/dsl/jobs.groovy
@@ -12,6 +12,10 @@ Map getMultijobPRConfig() {
             [
                 id: 'optaweb-vehicle-routing',
                 primary: true,
+                env : [
+                    // Disable Sonarcloud analysis.
+                    ENABLE_SONARCLOUD: false,
+                ]
             ]
         ],
     ]
--- a/.ci/jenkins/dsl/jobs.groovy
+++ b/.ci/jenkins/dsl/jobs.groovy
@@ -14,7 +14,7 @@ Map getMultijobPRConfig() {
                 primary: true,
                 env : [
                     // Disable Sonarcloud analysis.
-                    ENABLE_SONARCLOUD: false,
+                    DISABLE_SONARCLOUD: true                    
                 ]
             ]
         ],
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,8 @@
     <version.frontend-maven-plugin>1.11.3</version.frontend-maven-plugin>
     <version.node>v12.16.2</version.node>
     <version.npm>6.14.18</version.npm>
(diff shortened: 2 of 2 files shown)


mareknovotny:
jenkins retest this please

lucamolteni (approved):
I'm still convinced we should have been fine anyway with this https://github.com/quarkusio/quarkus/pull/30440 but I'm fine with this solution

yurloc:
> I'm still convinced we should have been fine anyway with this [quarkusio/quarkus#30440](https://github.com/quarkusio/quarkus/pull/30440) but I'm fine with this solution

@lucamolteni I don't think so because `smallrye-config-source-yaml:2.12.3` depends on `snakeyaml:1.33`, see https://mvnrepository.com/artifact/io.smallrye.config/smallrye-config-source-yaml/2.12.3. Am I missing something?

lucamolteni:
> > I'm still convinced we should have been fine anyway with this [quarkusio/quarkus#30440](https://github.com/quarkusio/quarkus/pull/30440) but I'm fine with this solution
> 
> @lucamolteni I don't think so because `smallrye-config-source-yaml:2.12.3` depends on `snakeyaml:1.33`, see https://mvnrepository.com/artifact/io.smallrye.config/smallrye-config-source-yaml/2.12.3. Am I missing something?

You're correct, it uses snakeyaml 1.33. We don't need to use snakeyaml 2.0.0 necessarily, as long as we don't use that specific no args constructor

yurloc:
> > > I'm still convinced we should have been fine anyway with this [quarkusio/quarkus#30440](https://github.com/quarkusio/quarkus/pull/30440) but I'm fine with this solution
> > 
> > 
> > @lucamolteni I don't think so because `smallrye-config-source-yaml:2.12.3` depends on `snakeyaml:1.33`, see https://mvnrepository.com/artifact/io.smallrye.config/smallrye-config-source-yaml/2.12.3. Am I missing something?
> 
> You're correct, it uses snakeyaml 1.33. We don't need to use snakeyaml 2.0.0 necessarily, as long as we don't use that specific no args constructor

Yes, that is true but our ultimate goal is not only to prevent the vulnerability (which is technically not present in this project) but also to escape the reporting tools, which mark the project as vulnerable just because it depends on SnakeYAML 1.33.

yurloc:
Jenkins retest this.

I want to check if https://github.com/kiegroup/optaplanner/pull/33 actually disables Sonar analysis.

yurloc (approved):
LGTM.

